PR #1, opened 2015-02-10
Encrypted cookie session

mind1master:


aiohttp_session/cookie_storage.py
```diff
@@ -33,3 +37,54 @@ def save_session(self, request, response):
...
+    """Encrypted JSON storage.
+    """
+
+    def __init__(self, secret_key, identity="AIOHTTP_COOKIE_SESSION", *,
+                 domain=None, max_age=None, path='/',
+                 secure=None, httponly=True, iv=None):
+        super().__init__(identity, domain=domain, max_age=max_age,
+                         path=path, secure=secure, httponly=httponly)
+
+        self.__secret_key = secret_key
```
asvetlov: Please don't use double underscore

aiohttp_session/cookie_storage.py
```diff
@@ -33,3 +37,54 @@ def save_session(self, request, response):
...
+        if len(self.__secret_key) % AES.block_size != 0:
+            raise TypeError(
+                'Secret key must be a multiple of {} in length'.format(
+                    AES.block_size))
+        self.__iv = iv
```
asvetlov: Initial vector should be never passed to storage constructor nor stored in storage attr. It should be stored in cookie together with encoded data.

asvetlov:
Thanks!

PR #7, opened 2015-04-19
stop cookie name growing each time session is saved

dannyclark:
I think this is a mistake: I end up with a longer key (one more 'cookie name' prefix) when I call get_session and this gets persisted to the browser if the session is saved

Also, I want to handle the case where the session in redis expires but the cookie is still present in the user's browser.


asvetlov:
Thanks for report. My fix is a bit different from your proposal but follows your idea.
I've released 0.1.1 for that fix.

dannyclark:
Works for me. Many thanks.

PR #11, opened 2015-05-30
Fernet from cryptography.io

iho:


gleb-chipiga:
Fernet uses small signing-key with 128-bit length for SHA-256 HMAC. This is strange decision

asvetlov:
Superseded by #16

Issue #4, opened 2015-04-01
Implement session.created attribute


asvetlov:
A timestamp indicating the time when this session was created.

asvetlov:
Done

PR #14, opened 2015-07-07
session.created property

GomZik:
Issue #4: Add created property with timestamp indicating the time when session was created


asvetlov:
At first I very appreciate your contribution.
But I think we should use a bit another data structure.
Instead of pushing `'created'` key into data dictionary we should create dictionary like:
```{'created': created_timestamp,
     'session': session_dict}
```
where current `data` moved into `session_dict`.
The benefits are:
1. we don't overlap session keys with our auxiliary ones. Say, user may want to have his own  `session['created']` value, why not?
2. I suspect `created` is not the only auxiliary session parameter. Say, we need to control session lifetime based on individual session info, not on system-wide value.

GomZik:
Thanks for your review. I will rewrite that feature.

GomZik:
Can you check my PR again, please?

aiohttp_session/__init__.py
```diff
@@ -20,8 +20,16 @@ def __init__(self, identity, *, data=None, new=False):
...
+        if new or created is None:
+            self._created = int(time.time())
+        else:
+            self._created = created
+
+        if session_data is not None:
+            self._mapping.update(session_data)
 
     def __repr__(self):
         return '<{} [new:{}, changed:{}] {!r}>'.format(
```
asvetlov: Please add `created` info to the repr.

aiohttp_session/__init__.py
```diff
@@ -36,6 +44,14 @@ def new(self):
     def identity(self):
         return self._identity
 
+    @property
+    def created(self):
+        return self._created
+
+    @property
+    def is_empty(self):
```
asvetlov: Please describe why do you need `is_empty` property.
What's your use case?

GomZik: Just a public property to determine is the session dict empty

aiohttp_session/__init__.py
GomZik: Ok, i'l do it

aiohttp_session/__init__.py
asvetlov: Ok than.
But name should be changed to `.empty` -- like `.changed` and `.created`

aiohttp_session/__init__.py
```diff
@@ -143,6 +159,16 @@ def max_age(self):
     def cookie_params(self):
         return self._cookie_params
 
+    def get_session_data(self, session):
```
asvetlov: I believe it's private method, it should be named `_get_session_data` (starting with underscore).

GomZik:
I made fixes from your comments

asvetlov:
Thanks!

PR #19, opened 2015-10-02
Add missing steps to the session redis storage example

avanov:
Current code snippet misses the ``aiohttp_session.session_middleware`` wrapper, which leads to AssertionError when one tries to copy-paste-run it.


asvetlov:
Thanks

PR #24, opened 2016-01-03
save session with HTTPException

imbolc:


asvetlov:
Please add `max_http_status` param to `setup()` also

imbolc:
Done, but I think, `setup()` is not so good idea. It's a second way to do the same. And it may cause obscure bugs because of order of middlewares

asvetlov:
Thanks. I think `setup()` is prime API for aiohttp plugins, and `setup()` calls may be ordered.
The reason is: `setup()` accepts application instance while middleware cannot accept it.

asvetlov:
Just after merging I've realized that `max_http_status` is a bad name.
`error_http_status` is slightly better.
@imbolc what do you think?

imbolc:
But it's not about errors only, eg `raise web.HTTPOk()`

asvetlov:
`failed_http_status` for first failed status code?

imbolc:
max_handled_status?

imbolc:
Or `max_exception_status` because it's about processing of exceptions only?

asvetlov:
Well, `max_http_status` maybe ok.
We just need good description for docs.
Is 'maximum HTTP status code for caught HTTPException which doesn't prevent session saving on request processing finish` sounds good?

imbolc:
Now i'm not sure it's good api at all :)

Maybe by default it should work with raising all of `HTTPExceptions`, because it works with returning this?
Additionally it may have setting like `process_exceptions`, eg:
```
process_exceptions=[web.HTTPException]  # by default
process_exceptions=[web.HTTPRedirection, web.HTTPSuccessful]  # do not works with errors
```

asvetlov:
Funny conclusion.
Let me take a pause and sleep on it.

imbolc:
Ok :) I meant if somebody raised `HTTPBadRequest` after changing of session. He apparently expect to  see this changing in the response. Is there any ideological differences between returning and raising of HTTPExceptions?

asvetlov:
I found returning HTTPException and raising it has the same meaning.
Downside of approach is requirement for middleware to process both in the same way, which may be  confusing for inexperienced developers.
We may force HTTPException raising by a small tweak in RequestHandler.handle_request logic though.

Anyway, for the library I'm inclined to save the session unconditionally.

If I change session state in web handler I expect the change will be saved regardless outer middleware codes.

Forbidding the rule may make a mess easy.

@imbolc do you agree?

imbolc:
Yes, I totally agree.

asvetlov:
Please make an updated PR -- I want to merge it.
We don't need any docs change as I see.
I've stuck on writing docs update for `max_http_status` since merging this one.

imbolc:
Done

PR #31, opened 2016-02-28
add encoder, decoder paramater

TechBK:
check my code. plzz told me if it wrong!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ install:
 - pip install aioredis
 - pip install cryptography
 - pip install pynacl
+- pip install pymongo
 - python setup.py develop
 
 script:
--- a/tests/test_cookie_storage_with_bson.py
+++ b/tests/test_cookie_storage_with_bson.py
@@ -1,0 +1,179 @@
+import asyncio
+import bson.json_util as json
+from bson import objectid
+import socket
+import unittest
+import time
+
+from aiohttp import web, request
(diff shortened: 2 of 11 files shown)


.travis.yml
```diff
@@ -14,6 +14,7 @@ install:
 - pip install aioredis
 - pip install cryptography
 - pip install pynacl
+- pip install pymongo
```
jettify: Please remove pymongo and bson altogether, there is no point to have it here. You should be able to test custom serializer without introducing this dependency.

tests/test_cookie_storage_with_bson.py
```diff
@@ -0,0 +1,179 @@
...
+import socket
+import unittest
+import time
+
+from aiohttp import web, request
+from aiohttp_session import (Session, session_middleware,
+                             get_session, SimpleCookieStorage)
+
+
+class TestSimleCookieStorage(unittest.TestCase):
```
jettify: Also no need to copy paste those files, just add simple test for each storage type that hits changed logic.

TechBK: Thanks for the feedback. this is my first time writing testcase.
I also thought like you but this is not completely copy / paste.
When I declared `from import bson json_util as json`, the creation of additional test file will make the test simpler and more explicit. Besides, I do not really understand the logic of the session changed. So I chose to copy / paste.
I'll fix it in the next commit.

PR #41, opened 2016-08-16
Added EncryptedRedisStorage

bcmyers:
Hello.  I've never tried a pull request before; so I hope I'm doing this right.  

I don't know if it's useful, but I added another storage class: EncryptedRedisStorage.  Instead of just storing a UUID on the client (like the RedisStorage class does), it stores an encrypted UUID on the client.

I'm not sure if this "extra security" is useful, because it seems pretty safe to store a UUID on the client, but maybe it is.  If you think this is something that might help, I'm glad I could contribute.  If not, no worries.  Love the library!

### DIFF
--- a/aiohttp_session/redis_storage.py
+++ b/aiohttp_session/redis_storage.py
@@ -2,6 +2,9 @@ import asyncio
 import json
 import uuid
 
+from cryptography import fernet
+from cryptography.fernet import InvalidToken
+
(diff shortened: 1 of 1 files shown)


asvetlov:
Sorry, but UUID encrypting doesn't make sense.
The patch reduces performance without increasing protection level.

Issue #42, opened 2016-08-27
Set expires along with max_age


asvetlov:
See KeepSafe/aiohttp#1056

asvetlov:
Fixed by #43

PR #43, opened 2016-09-07
Add expires field automatically to support older browsers

artyomboyko:
Fixing #42

### DIFF
--- a/aiohttp_session/__init__.py
+++ b/aiohttp_session/__init__.py
@@ -210,7 +210,8 @@ class AbstractStorage(metaclass=abc.ABCMeta):
         params = dict(self._cookie_params)
         if max_age is not None:
             params['max_age'] = max_age
-            params['expires'] = time.strftime("%a, %d-%b-%Y %T GMT", time.gmtime(time.time() + max_age))
+            params['expires'] = time.strftime("%a, %d-%b-%Y %T GMT",
+                                              time.gmtime(time.time() + max_age))
         if not cookie_data:
             response.del_cookie(self._cookie_name)
         else:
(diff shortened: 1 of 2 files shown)


aiohttp_session/__init__.py
```diff
@@ -210,6 +210,7 @@ def save_cookie(self, response, cookie_data, *, max_age=None):
         params = dict(self._cookie_params)
         if max_age is not None:
             params['max_age'] = max_age
+            params['expires'] = time.strftime("%a, %d-%b-%Y %T GMT", time.gmtime(time.time() + max_age))
```
asvetlov: `loop.time()` should be used

asvetlov:
Please add a test.

aiohttp_session/__init__.py
artyomboyko: Sorry for stupid question, does that mean
```
self._loop = loop if loop is not None else asyncio.get_event_loop()
``` 
? Because communication between `aiohttp` and `aiohttp_session` is very limited, don't see other way.

asvetlov: I've thought the storage already has a loop.
Nevermind.
But please replace `time.time()` with `time.monotomic()`.

tests/test_abstract_storage.py
```diff
@@ -0,0 +1,46 @@
...
+def test_max_age_also_returns_expires(test_client):
+
+    @asyncio.coroutine
+    def handler(request):
+        time.monotonic.return_value = 0.0
+        session = yield from get_session(request)
+        session['c'] = 3
+        return web.Response(body=b'OK')
+
+    with mock.patch('time.monotonic') as MockClass:
```
asvetlov: Please use `m_monotonic` instead of `MockClass`.
Your patch is perfectly fine but all my projects use `m_*` naming for mocking.
I'd love to keep naming rules consistent.

asvetlov:
Thank you very much!

Issue #44, opened 2016-09-07
Respect session.max_age in redis storage


asvetlov:
Here https://github.com/aio-libs/aiohttp_session/blob/master/aiohttp_session/redis_storage.py#L51
should be `max_age=session.max_age`.

asvetlov:
Fixed by #45

PR #45, opened 2016-09-07
Respect session.max_age in redis storage

artyomboyko:
#44

### DIFF
--- a/aiohttp_session/__init__.py
+++ b/aiohttp_session/__init__.py
@@ -16,12 +16,12 @@ class Session(MutableMapping):
 
     """Session dict-like object."""
 
-    def __init__(self, identity, *, data, new):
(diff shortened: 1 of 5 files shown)


artyomboyko:
Didn't find a better way to set expire date in the redis ...

artyomboyko:
@asvetlov  Not sure how to fix it, looks like session.max_age is always empty. Not sure where it's supposed to be set.

asvetlov:
Yes, it is.
The main idea is: use `storage.max_age` as default value for `session.max_age` but allow to user overriding the last value in web-handler on per-user basis.

artyomboyko:
@asvetlov what do you think ? looks like it's done. did I understood you correctly ?

asvetlov:
Cool!

Issue #85, opened 2016-12-12
Clarify the Redis key inside the docs


barrachri:
Actually the doc doesn't specify that they key inside Redis would be: AIOHTTP_SESSION_valueofthecookie

asvetlov:
Who need this info?
Anyway, PR is welcome!

barrachri:
Well my use case is probably "strange" but I needed to check the validity of the session outside aiohttp.

I will send a PR !

jettify:
Resolved via https://github.com/aio-libs/aiohttp-session/pull/94

PR #94, opened 2017-01-15
add info about redis key inside the doc

barrachri:
https://github.com/aio-libs/aiohttp-session/issues/85


jettify:
Thanks a lot!

PR #102, opened 2017-01-24
Add some labels

alxpy:
/cc @jettify


jettify:
could you also add this one too:
```
.. image:: https://img.shields.io/pypi/v/aiohttp-session.svg
    :target: https://pypi.python.org/pypi/aiohttp-session
```

alxpy:
@jettify done

jettify:
Thanks!

PR #103, opened 2017-01-25
properly delete a cookie with path or domain set

sobczyk:
session was not deleted properly, and persisted when invalidate was called.
I detected it on subdomain cookies as domain=".example.com"


jettify:
Could you please add test case? and fix flake8 error?

sobczyk:
I'm not sure how to fake the domain name when making requests to the server.
I can add a Host header manually, but the client is making requests to localhost, and overwrites it anyway. Moreover it's dependent on client's CookieJar host matching

fafhrd91:
@asvetlov @jettify ping

asvetlov:
Superseded by #171

PR #166, opened 2017-05-23
[Redis] Return expired cookie on session.invalidate

panagiks:
and test cases.

### DIFF
--- a/aiohttp_session/redis_storage.py
+++ b/aiohttp_session/redis_storage.py
@@ -44,7 +44,7 @@ class RedisStorage(AbstractStorage):
                              max_age=session.max_age)
         else:
             if session.empty:
-                self.save_cookie(response, session._mapping,
+                self.save_cookie(response, '',
                                  max_age=session.max_age)
             else:
                 key = str(key)
--- a/aiohttp_session/redis_storage.py
+++ b/aiohttp_session/redis_storage.py
@@ -46,9 +46,10 @@ class RedisStorage(AbstractStorage):
             if session.empty:
                 self.save_cookie(response, session._mapping,
                                  max_age=session.max_age)
-            key = str(key)
-            self.save_cookie(response, key,
-                             max_age=session.max_age)
+            else:
+                key = str(key)
+                self.save_cookie(response, key,
+                                 max_age=session.max_age)
 
         data = self._encoder(self._get_session_data(session))
         with (yield from self._redis) as conn:
--- a/aiohttp_session/redis_storage.py
+++ b/aiohttp_session/redis_storage.py
@@ -43,6 +43,9 @@ class RedisStorage(AbstractStorage):
             self.save_cookie(response, key,
                              max_age=session.max_age)
         else:
+            if session.empty:
+                return self.save_cookie(response, session._mapping,
+                                        max_age=session.max_age)
             key = str(key)
             self.save_cookie(response, key,
                              max_age=session.max_age)
--- a/aiohttp_session/redis_storage.py
+++ b/aiohttp_session/redis_storage.py
@@ -44,8 +44,8 @@ class RedisStorage(AbstractStorage):
                              max_age=session.max_age)
         else:
             if session.empty:
-                return self.save_cookie(response, session._mapping,
-                                        max_age=session.max_age)
+                self.save_cookie(response, session._mapping,
+                                 max_age=session.max_age)
             key = str(key)
             self.save_cookie(response, key,
                              max_age=session.max_age)
--- a/tests/test_redis_storage.py
+++ b/tests/test_redis_storage.py
@@ -143,8 +143,9 @@ def test_clear_cookie_on_sesssion_invalidation(test_client, redis):
     value = yield from load_cookie(client, redis)
     assert {} == value
     morsel = resp.cookies['AIOHTTP_SESSION']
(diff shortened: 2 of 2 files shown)


panagiks:
With the current structure of the test file I can't seem to be able to test expiration. That is because max_age seems to be set at None for the tests. Because of this, and because of [these lines](https://github.com/aio-libs/aiohttp-session/blob/master/aiohttp_session/__init__.py#L211-L215) the "expires" and the "max-age" fields are empty strings (visible in the failed travis tests).

If someone else can have a look at the tests (the test cases are available in my previous commit) ...

panagiks:
(Relates to #163 )

panagiks:
Hmmm it doesn't seem to be caused by the max_age=None .... I've written this helper script to see what the cookie contains : 

```py
import asyncio
import aiohttp

@asyncio.coroutine
def do_sth():
    ses = aiohttp.ClientSession()
    me = {mylogin}
    r = yield from ses.post('http://127.0.0.1:9045/login/', json=me)
    ses = aiohttp.ClientSession(cookies=r.cookies)
    r = yield from ses.get('http://127.0.0.1:9045/logout/')
    print(r.cookies['MYCOOKIE']['expires']) # prints "Thu, 01 Jan 1970 00:00:00 GMT"
    print(r.cookies['MYCOOKIE']['max-age']) # prints 0

def main():
    loop = asyncio.get_event_loop()
    loop.run_until_complete(do_sth())

if __name__ == "__main__":
    main()
```
and my logout function is
```py
@asyncio.coroutine
def logout(request):
    session = yield from get_session(request)
    if not session.new:
        session.invalidate()
    return json_response('', status=200)
```

But `expires` in the `morsel` variable used in the tests is '' and so is `max-age`

panagiks:
Ok I found it! Looking at [aiohttp's source for del_cookie](http://aiohttp.readthedocs.io/en/stable/_modules/aiohttp/web_response.html#StreamResponse.del_cookie) I found that it never sets `httponly` so [this test](https://github.com/aio-libs/aiohttp-session/blob/master/tests/test_redis_storage.py#L146) actually guarantees that the cookie was NOT deleted. After realizing this I changed another thing to the code so it makes more sense functionally and removed the test for `httponly` during session invalidation for Redis and also added back the rest of the tests. It should be good now pushing soon.

panagiks:
Hi @asvetlov, @jettify,

Is there anything else that has to be done? Or any objection in the way this was implemented? If there is just let me know so I can fix it :)

panagiks:
Hey, sorry for bumping this but it's quite bumming to not be able essentially logout my api's users ... is there any addition/correction on this PR you need from me to be able to merge it?

asvetlov:
Sorry, I'm on trip.
Will review the PR in a week.

aiohttp_session/redis_storage.py
```diff
@@ -43,9 +43,13 @@ def save_session(self, request, response, session):
             self.save_cookie(response, key,
                              max_age=session.max_age)
         else:
-            key = str(key)
-            self.save_cookie(response, key,
-                             max_age=session.max_age)
+            if session.empty:
+                self.save_cookie(response, session._mapping,
```
jettify: I think `session` should be explicitly converted to dict, because `_mapping` is private and implementation detail.

jettify: bit confused, `session._mapping` is empty dict right?  I think ` self.save_cookie` expects string there.

panagiks: Yes. You are correct (about both). Basically anything passed in `cookie_data` field that evaluates to false will result in a deletion of the cookie so passing `''` instead of `session._mapping` would have the same result and would also be consistent usage of the function but I didn't like it (from an appearance perspective), but if you prefer passing an empty string there I have no practical objection to it. Just drop a comment here and I'll push another commit to change it :) Also thank you for looking into it :)

panagiks: Ah yes now I remembered why I used session._mapping specifically :P It's what is already done in `cookie_storage.py` to delete the cookie (see [here](https://github.com/aio-libs/aiohttp-session/blob/master/aiohttp_session/cookie_storage.py#L48-L50))

jettify: Got it, @asvetlov do you remember anything about:
```
        if session.empty:
            return self.save_cookie(response, session._mapping,
                                    max_age=session.max_age)
``` 
??

panagiks: It seems to have been added by [this commit ](https://github.com/aio-libs/aiohttp-session/commit/d8b8dd619ecd801f77fcf5bdeb506aca01e3ea14).

jettify: I see, lets give @asvetlov some time to take a look. As for me I think we can merge it.

asvetlov: I think it should be empty string in both redis and cookie storage

panagiks: Is it ok if I push the change here for redis and open a new pr for cookie storage (so that the prs are thematically correct)?

asvetlov: Ok

panagiks:
Ok done :)

jettify:
thanks!

Issue #28, opened 2016-01-27
Q: uuid4 vs urandom(16) for Redis keys


jashandeep-sohi:
This is more of question than an issue (and maybe not even specific to `aiohttp_session`), but I was wondering why are the keys using `uuid4` when `codecs.encode(urandom(16), 'hex')` would suffice?
Is it just a preference or is there some deeper reason?

asvetlov:
It's just my preference.
uuid4 uses `urandom(16)` internally.

jashandeep-sohi:
Would it be okay if I add a `key_factory` parameter to `RedisStorage`? Default could still be `uuid4`.

asvetlov:
Please do

smialy:
#205

smialy:
To close?

asvetlov:
Fixed by #205

PR #205, opened 2017-09-12
Add key_factory attr for redis_storage

smialy:
Fix for #28


jettify:
I thinks it is good idea, could you please add test case with custom key factory?

smialy:
Not problem.

jettify:
Thanks! will merge tonight

asvetlov:
@jettify please make a new release after merging

jettify:
New version on pypi https://pypi.python.org/pypi/aiohttp_session/1.0.1

PR #208, opened 2017-09-19
cleaned up documentation, fixed demo to work with a recent version of aiohttp

dfee:


demo/main.py
```diff
@@ -1,24 +1,27 @@
...
-def handler(request):
-    session = yield from get_session(request)
+
+async def handler(request):
+    session = await get_session(request)
     last_visit = session['last_visit'] if 'last_visit' in session else None
     session['last_visit'] = time.time()
     text = 'Last visited: {}'.format(last_visit)
-    return web.Response(body=text.encode('utf-8'))
+    return web.Response(body=text)
```
asvetlov: No. Either `body=text.encode()` or `text=text`

dfee: Just saw this and submitted a change. Sorry about that!

dfee: Firefox kept downloading the file, and I couldn't figure out why... I had to set `content_type='text/html'` or use `text=` as it sets a nice content_type for us.

PR #229, opened 2017-11-12
Use Docker to test external storages 

### LINKED ISSUE #226: Rewrite tests to use docker
Like aiopg
We need to use dockerized redis and memcache servers in our tests.
It simplifies local development.

panagiks:
Fixes #226.

Both memcached and redis are now tested using Docker.

Having said that there were a couple of things I couldn't get to work:
* Using aiodocker (as @jettify suggested), sorry :s I also liked the idea of using aiodocker
* Using tags for the containers (now the latest image is pulled), pytest wasn't recognizing the `redis_tag` (that I had created using aiopg tests as an example) as a pytest fixture.

I believe that both can be done in the future as an enhancement, but the decision is up to you.


tests/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+import aioredis
+import asyncio
+import gc
+import pytest
+import time
+import uuid
+from docker import Client as DockerClient
+
+
+@pytest.yield_fixture(scope='session')
```
asvetlov: I pretty sure `yield_fixture` could be replaced by just `@pytest.fixture` in modern pytest.
We still have `yield_fixture` in very many *aio-libs* projects.
Need for a hero for replace them all or ending of deprecation period for `@pytest.fixture` :)

panagiks: I can go through a few of them (probably next weekend) :)

tests/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+        loop.call_soon(loop.stop)
+        loop.run_forever()
+        loop.close()
+    gc.collect()
+    asyncio.set_event_loop(None)
+
+
+@pytest.fixture(scope='session')
+def session_id():
+    '''Unique session identifier, random string.'''
```
jettify: Very minor thing, lets use double quotes here.

panagiks: Sure :)

tests/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+        detach=True,
+    )
+    docker.start(container=container['Id'])
+    inspection = docker.inspect_container(container['Id'])
+    host = inspection['NetworkSettings']['IPAddress']
+    delay = 0.001
+    for i in range(100):
+        try:
+            conn = loop.run_until_complete(
+                aioredis.create_connection((host, 6379))
```
jettify: in order to fix 
```
RuntimeError: There is no current event loop in thread 'MainThread'.
```
you need to pass loop explicitly here

jettify:
PR lock good to me, we just need to make it green

tests/conftest.py
panagiks: To avoid multiple pushes, you mean pass it to `run_until_complete` as a kwarg right?

panagiks: (retracted run_until_complete doesn't take loop kwarg :s so I guess in create_connection)

panagiks:
Should be ok now :)

jettify:
@panagiks thanks a lot!

asvetlov:
Awesome work!

panagiks:
Thank you :)

PR #234, opened 2017-11-20
New async/await syntax and aioredis 1.0.0 compatibility

panagiks:
Drops support for Python<3.5.3
Drops support for aioredis<1.0.0

Caveats:
* The reason support for aioredis<1.0.0 was dropped:
  * create_pool returns a `ConnectionsPool` object
  * when `ConnectionsPool` is used with context management returns a `RedisConnection` object
  * `RedisConnection` doesn't have `set` / `get` methods in v1.0.0+
  * create_redis_pool (introduced in v1.0.0 ) returns a `Redis` object that wrapps `ConnectionsPool`
  * when `Redis` is used whith context management returns a `ContextRedis` object that has `set` / `get`

My initial idea was to check whether the supplied pool was `ConnectionsPool` or `Redis` and in the first case wrap it:
```py
if isinstance(redis_pool, aioredis.pool.ConnectionsPool):
    redis_pool = aioredis.commands.Redis(redis_pool)
elif not isinstance(redis_pool, aioredis.commands.Redis):
    raise something
```

But I was not sure about it. If you prefer the above over straight up asserting that `redis_pool` is instance of `Redis` I can add it.

Finally there is the issue that users of `aiohttp-session` will now have to generate the redis_pool that they provide to `RedisStorage` using `create_redis_pool` instead of `create_pool` and I'm not sure what the best approach to communicate this to them would be.

### DIFF
--- a/aiohttp_session/redis_storage.py
+++ b/aiohttp_session/redis_storage.py
@@ -1,7 +1,11 @@
-import aioredis
+try:
+    import aioredis
+except ImportError:
+    aioredis = None
 import json
 import uuid
 
+from distutils.version import StrictVersion
 from . import AbstractStorage, Session
 
 
@@ -16,6 +20,10 @@ class RedisStorage(AbstractStorage):
         super().__init__(cookie_name=cookie_name, domain=domain,
                          max_age=max_age, path=path, secure=secure,
                          httponly=httponly)
+        if aioredis is None:
+            raise RuntimeError("Please install aioredis")
+        if StrictVersion(aioredis.__version__).version < (1, 0):
+            raise RuntimeError("aioredis<1.0 is not supported")
         self._encoder = encoder
         self._decoder = decoder
         self._key_factory = key_factory
--- a/aiohttp_session/redis_storage.py
+++ b/aiohttp_session/redis_storage.py
@@ -1,6 +1,6 @@
 try:
     import aioredis
-except ImportError:
+except ImportError:  # pragma: no cover
     aioredis = None
 import json
 import uuid
--- a/aiohttp_session/redis_storage.py
+++ b/aiohttp_session/redis_storage.py
@@ -4,6 +4,7 @@ except ImportError:
     aioredis = None
 import json
 import uuid
+import warnings
 
 from distutils.version import StrictVersion
 from . import AbstractStorage, Session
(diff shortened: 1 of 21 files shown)


asvetlov:
Would you update `.travis.yml` to check on Python 3.6 too?

panagiks:
Sure :)

aiohttp_session/redis_storage.py
```diff
@@ -1,4 +1,4 @@
-import asyncio
+import aioredis
```
asvetlov: `aioredis` is an **optional** dependency.

Please use the following:
```
try:
     import aioredis
except ImportError:
    aioredis = None
```

In `RedisStorage.__init__` add lines:
```
if aioredis is None:
    raise RuntimeError("Please install aioredis")
```
Also I suggest to add aioredis version check because `aioredis` is an optional dependency and the version cannot be checked by `pip`:
```
from distutils.version import StrictVersion  # aioredis supports strict version schema
if StrictVersion(aioredis.__version__).version < (1, 0):
    raise RuntimeError("aioredis<1.0 is not supported")
```
This explicitly points *user* to the problem.

aiohttp_session/redis_storage.py
```diff
@@ -19,17 +19,17 @@ def __init__(self, redis_pool, *, cookie_name="AIOHTTP_SESSION",
         self._encoder = encoder
         self._decoder = decoder
         self._key_factory = key_factory
+        assert isinstance(redis_pool, aioredis.commands.Redis), redis_pool
```
asvetlov: Let's left `assert` and `AssertionError` to unittests, explicit `TypeError` is much better.

asvetlov:
Regarding to `RuntimeError` -- I don't sure but cannot choose better exception.
About `assert` statements in the library body -- they all an artifact from very early development stages.

If you'll replace them by explicit error raising -- I'll very appreciate.
But please do it in other PR.

aiohttp_session/redis_storage.py
panagiks: Would you agree if I replaced this with:
```py
if isinstance(redis_pool, aioredis.pool.ConnectionsPool):
    warnings.warn(
        "using a pool created with aioredis.create_pool is deprecated"
        "please provide a pool created with aioredis.create_redis_pool",
        DeprecationWarning
    )
    redis_pool = aioredis.commands.Redis(redis_pool)
elif not isinstance(redis_pool, aioredis.commands.Redis):
    raise TypeError("Expexted aioredis.commands.Redis got {}".format(type(redis_pool)))
```

This way we keep compatibility with the old `create_pool` function, allowing for a smoother transition of the users. (It should work from a look I had in `aioredis` source but to be honest I haven't tested it yet)

panagiks: I'll include the above changes in my next commit. If you disagree with that implementation / approach feel free to drop me a line and I'll revert them :)

panagiks:
Can't find if/how to cover the initial import of `aioredis` with tests :s (The only uncovered change)

asvetlov:
Add `# pragma: no cover` comment to `except ImportError` line.
Technically it's possible to make another test run without `aioredis` installed -- but it doesn't make much sense and increases tests execution a lot.

asvetlov:
Thanks!

asvetlov:
2.0.0 released

PR #237, opened 2017-11-24
Method set_new_identity for change identity

### LINKED ISSUE #236: Add the ability set `identity` of session if it is `new`
Hello,

What do you think about it?

I need set `identity`, which used how `key` in redis,  of session with custom key. And then I need remove session by that `identity` from redis by message from amqp with this `identity`.

Now `identity` [readonly property](https://github.com/aio-libs/aiohttp-session/blob/master/aiohttp_session/__init__.py#L45).

Example code
```python
class LoginView(BaseView):
   async def post(self):
       session = await aiohttp_session.get_session(self.request)
       if session.new:
           session.identity = 'key'
       session['user'] = {...}
```

Thanks.

achernetsky:
resolve #236

### DIFF
--- a/aiohttp_session/__init__.py
+++ b/aiohttp_session/__init__.py
@@ -68,6 +68,13 @@ class Session(MutableMapping):
         self._changed = True
         self._mapping = {}
 
+    def set_new_identity(self, identity):
+        if not self._new:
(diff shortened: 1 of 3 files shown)


asvetlov:
Perfect!

PR #252, opened 2018-02-07
Move encoder and decoder to AbstractStorage

BFriedrichs:
I left the args explicit, means that json still needs to be imported in all files. The first solution using kwargs seemed kind of odd considering we're passing through everything else explicitly.

I didn't really look into it much but I couldn't get all tests to run because docker seems to be having hiccups at some point. Not sure if MacOS might be the culprit here.

About the docs, :class:`func` doesn't seem right but I'm new to Sphinx docs and couldn't get :func: parameters to work.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -180,8 +180,8 @@ implement both :meth:`~AbstractStorage.load_session` and
    A callable with the following signature: `def encode(param: Any) -> str: ...`.
    Default is :func:`json.dumps`.
 
-   *decoder* -- session deserializer. A callable with the following signature:
-   `def decode(param: str) -> Any: ...`.
+   *decoder* -- session deserializer.
+   A callable with the following signature: `def decode(param: str) -> Any: ...`.
    Default is :func:`json.loads`.
 
    .. attribute:: max_age
@@ -200,15 +200,15 @@ implement both :meth:`~AbstractStorage.load_session` and
 
    .. attribute:: encoder
 
-      .. versionadded:: 2.3.0 
-
       The JSON serializer that will be used to dump session cookie data.
+
+      .. versionadded:: 2.3
 
    .. attribute:: decoder
 
-      .. versionadded:: 2.3.0 
-      
       The JSON deserializer that will be used to load session cookie data.
+
+      .. versionadded:: 2.3
 
    .. method:: load_session(request)
 
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -176,9 +176,13 @@ implement both :meth:`~AbstractStorage.load_session` and
    *httponly* -- cookie's http-only flag, :class:`bool` or ``None`` (the
    same as ``False``).
 
-   *encoder* -- json serializer, :class:`func`
+   *encoder* -- session serializer.
+   A callable with the following signature: `def encode(param: Any) -> str: ...`.
+   Default is :func:`json.dumps`.
 
-   *decoder* -- json deserializer, :class:`func`
+   *decoder* -- session deserializer. A callable with the following signature:
+   `def decode(param: str) -> Any: ...`.
+   Default is :func:`json.loads`.
 
    .. attribute:: max_age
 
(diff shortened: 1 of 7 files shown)


docs/reference.rst
```diff
@@ -175,6 +176,10 @@ implement both :meth:`~AbstractStorage.load_session` and
    *httponly* -- cookie's http-only flag, :class:`bool` or ``None`` (the
    same as ``False``).
 
+   *encoder* -- json serializer, :class:`func`
+
+   *decoder* -- json deserializer, :class:`func`
```
asvetlov: Please use `versionadded`: http://www.sphinx-doc.org/en/stable/markup/para.html#directive-versionadded and specify required signatures for encoder/decoder

asvetlov:
Explicit parameters are perfect.

What problem with tests do you have?

docs/reference.rst
BFriedrichs: Would we bump the version first? I'm not entirely sure what the correct way to specify signatures in sphinx is. :func:`encoder(obj)` but that just generates `encoder(obj)()`.
I haven't really used sphinx before so I'm most certainly missing something.

BFriedrichs:
`ERROR at setup of test_create_new_sesssion`, followed by a giant list of, what I can only assume are, tracebacks in various packages.

Ending in 
`docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))`

docs/reference.rst
asvetlov: Next version will be 2.3, please use this number.

Replace :class:\`func\` with 
```
*encoder* -- session serializer.
A callable with the following signature: `def encode(param: Any) -> str: ...`.
Default is :func:`json.dumps`.

```

asvetlov:
Tests require docker server installed.

BFriedrichs:
The changes should be done, however I'm still unable to test Redis and Memcached because I think it tries to pull the wrong address here.
```
inspection = docker.inspect_container(container['Id'])
host = inspection['NetworkSettings']['IPAddress']
```

docs/reference.rst
```diff
@@ -189,6 +198,18 @@ implement both :meth:`~AbstractStorage.load_session` and
       :class:`dict` of cookie params: *domain*, *max_age*, *path*,
       *secure* and *httponly*.
 
+   .. attribute:: encoder
+
+      .. versionadded:: 2.3.0
```
asvetlov: Replace 2.3.0 with 2.3

asvetlov: Move versionadded below next line

docs/reference.rst
```diff
@@ -189,6 +198,18 @@ implement both :meth:`~AbstractStorage.load_session` and
...
 
+   .. attribute:: encoder
+
+      .. versionadded:: 2.3.0 
+
+      The JSON serializer that will be used to dump session cookie data.
+
+   .. attribute:: decoder
+
+      .. versionadded:: 2.3.0
```
asvetlov: The same

docs/reference.rst
```diff
@@ -175,6 +176,14 @@ implement both :meth:`~AbstractStorage.load_session` and
...
    same as ``False``).
 
+   *encoder* -- session serializer.
+   A callable with the following signature: `def encode(param: Any) -> str: ...`.
+   Default is :func:`json.dumps`.
+
+   *decoder* -- session deserializer. A callable with the following signature:
+   `def decode(param: str) -> Any: ...`.
+   Default is :func:`json.loads`.
+
```
asvetlov:
```
.. versionadded: 2.3
    Support *encoder* and *decoder* parameters.

BFriedrichs:
````
   *encoder* -- session serializer.
   A callable with the following signature: `def encode(param: Any) -> str: ...`.
   Default is :func:`json.dumps`.

   *decoder* -- session deserializer.
   A callable with the following signature: `def decode(param: str) -> Any: ...`.
   Default is :func:`json.loads`.

   .. versionadded: 2.3
      Support *encoder* and *decoder* parameters.
````
Adding it like so doesn't seem to have any impact on the doc generation.
It seems like `versionadded` cannot just be used inside of a class description like this?

jettify:
Docker is easy to fix for macOS, see example here:
https://github.com/aio-libs/aiopg/blob/3b670eaec23d4b2c3c024f9b8b82af02986afb6c/tests/conftest.py#L139-L145

Basically docker on Mac does not allow access to internal docker ip address by default, so you need to expose port to localhost.

BFriedrichs:
Thanks for your help @jettify. So I adjusted the tests for macOS in the latest commit. Using those I didn't come across any issues. Are any further additions / changes required?

asvetlov:
Everything is fine, thanks!

asvetlov:
I'll publish next release after making tests passing on aiohttp 3.0

PR #267, opened 2018-04-25
Added Postgresql storage

lysyloren:
Postgresql storage with asyncpg and aiopg support.


asvetlov:
Thanks for the contribution.
Unfortunately, I'm going to decline the pull requests after careful thinking.
There are several reasons for this:
1. Storing a temporary data like session info into SQL database is anti-pattern. Encouraging it *officially* doesn't sound like a good idea.
2. There are too many ways to store session data in database. JSONB is not the best format, most likely every concrete case will add application specific database columns and other constraints.

I'm ok to keep this code in examples for a reference. If you want -- feel free to move everything into (not existing yet) `examples` folder.

lysyloren:
OK, I understand that **by design** you do not want to support SQL database in aiohttp-session. But your arguments are understandable.

> Storing a temporary data like session info into SQL database is an anti-pattern. Encouraging it officially doesn't sound like a good idea.

Django, Pyramid, Rails, ASP.NET MVC (and many more), all of this frameworks **do** store session data in SQL database (as one of many supported options). Are these frameworks anti-patterns? Can you extend your idea of not storing session data in SQL database? Modern SQL RDBMSs can be used with either permanent or temporary data successfully. Please do not follow blindly with NoSQL trend, rejecting existing solid tools. End user will choose the best option for his deployment.

> There are too many ways to store session data in a database. JSONB is not the best format

JSONB in my code is only **optional feature** for advanced usage. Default storage method is json encoded TEXT column.

> most likely every concrete case will add application-specific database columns and other constraints.

Sure, that's why PostgresqlStorage can be easily integrated with almost any existing user SQL tables, by passing  any of the paramaters: `schema_name, table_name, column_name_key, column_name_data, column_name_expire, data_type`.

asvetlov:
Anyway the solution which dictates a table structure is not very generic.
Please consider my objections like "guts feeling".
If you want to use RDBMS -- feel free to do it.
If you want to reuse implementation -- just create a separate project for it.
A custom storage doesn't require any `aiohttp-session` modification.

asvetlov:
Yes, sounds like a good plan.

lysyloren:
Yes, I'll do. Thanks.

PR #269, opened 2018-04-28
Add login required example

bmwant:
## Rationale

* Add a hint to readme of how to create dev environment for the contributors.
* Add an example of how to implement `login_required` decorator in `aiohttp` web application elevating `aiohttp-session` package.


bmwant:
@asvetlov @jettify I think this example can be used as a reference when implementing such a thing with a help of `aiohttp-session` package. Do you think we can add this example?

Issue #25, opened 2016-01-03
Flash messages support


jettify:
Do we need some sort of flash messages support? Or may be this feature is out of scope of aiohttp_session?

Most frameworks have flash messages:
pyramid: http://docs.pylonsproject.org/projects/pyramid/en/1.0-branch/narr/sessions.html#flash-messages
flask: http://flask.pocoo.org/docs/0.10/patterns/flashing/

Particularly pyramid has it integrated to session object.

imbolc:
I think it will look better as a separate package because it brings some dependencies like `aiohttp_jinja2` for context_processor implementation.

jettify:
Do not follow, it is couple of methods:
`flash()` - add message to the queue, where queue is located in session object
`pop_flash()` - pop messages from the queue

Do not see why it should depend on `aiohttp_jinja2`

imbolc:
Because flash messages doesn't make much sense without templates. And I think there is not so much code to split it into parts. Exactly it took 22 loc in my last implementation: https://gist.github.com/imbolc/ca43545ca31329b4f9e2

imbolc:
Also, this approach has benefits of brevity:
```
flash(request, ('error', 'Foo error'))
```
instead of 
```
session = await get_session(request)
session.flash(request, ('error', 'Foo error'))
```

asvetlov:
@imbolc `request` is not necessary in your last example. `session` is coupled to request itself.
Anyway, let's postpone the issue.
For now I have a feeling that flash messages are part of UI library (while they are built on top of user session, sure).

Another interesting question is API for CSRF checks BTW.

shicky:
I also do not understand how flash depends on template engine.
As @asvetlov mentioned, flash is necessary to implement CSRF check.

IlyaSemenov:
I created the simple session flash library based on @imbolc's gist: 

https://github.com/IlyaSemenov/aiohttp_session_flash

It has a few enhancements over the original gist:

* session is not re-saved on each HTTP request unless flash messages were pushed or popped
* it doesn't lose flash messages followed by `raise web.HTTPFound(...)`
* it has tests and pypi package

TensorTom:
I did not know about the flash example and I implemented my own thing. I didn't use middleware though. I did this:
```
class Session:
    def __init__(self, session):
        self.session = session

    async def get(self, getter):
        return self.session[getter] if getter in self.session else None

    async def set(self, setter, content):
        self.session[setter] = content
```
[...]

PR #270, opened 2018-04-29
Flash messages example

bmwant:
## Rationale

We have an issue https://github.com/aio-libs/aiohttp-session/issues/25 for message flashing. I'm not sure what will be the final decision on that but we can at least include the snippet to the examples for users to refer when implementing such a feature in their application.

### DIFF
--- a/demo/flash_messages_example.py
+++ b/demo/flash_messages_example.py
@@ -1,0 +1,54 @@
+import base64
+from cryptography import fernet
+from aiohttp import web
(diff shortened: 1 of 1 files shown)


bmwant:
@asvetlov @jettify do you think we can include this example to the package?

asvetlov:
Thanks for the contribution.
I'm on the conference now, will review in a few days.

bmwant:
@asvetlov I know, listening you talk atm))

PR #273, opened 2018-04-30
Patch Session Fixation in Redis Storage

### LINKED ISSUE #272: Session Fixation vulnerability in RedisStorage
There is a window of opportunity for [Session Fixation](https://www.owasp.org/index.php/Session_fixation) exploitation in the logic of RedisStorage.

As seen here: https://github.com/aio-libs/aiohttp-session/blob/master/aiohttp_session/__init__.py#L190

Get session data returns an empty dictionary for an empty (this includes invalidated) session.

Referring here: https://github.com/aio-libs/aiohttp-session/blob/master/aiohttp_session/redis_storage.py#L60

save_session takes this data and saves it in Redis.

As a result, an invalidated session will result to the session ID being present in Redis with an empty mapping as its value.

Now looking over at: https://github.com/aio-libs/aiohttp-session/blob/master/aiohttp_session/redis_storage.py#L50

RedisStorage's load_session only looks at the case where data (returned by reading from Redis) is None. This will happen only if the key (session ID) is not present in Redis (has either expired or was never inserted) but as we established above the key is never actually removed, just the value mapping emptied. As a result the load_session function will return a session with the presented session ID and not a new one, although there was no valid session in storage for this ID.

If this is not caught and mitigated by the web app the following scenario can unfold:

* Attacker acquires a valid cookie
* Invalidates it (logs out)
* Attacker injects said cookie in victim's browser (see OWASP's link above for examples on how)
* Victim visits web app presenting the cookie present in his browser
* Web app uses the get_session to get a session object for the user, expecting a 'clean' session
* get_session returns a session with the session ID that was present in the cookie presented by the user
* session is populated by the web app and subsequently stored by aiohttp-session during the response
* User is now logged in with the session ID of the cookie that was injected by the attacker
* The attacker now controls (knows) a session cookie for a given user

panagiks:
Fixes #272 .

This is a relatively naive approach on patching the issue and there probably should be a more general/broad look into it design wise at some point. 

The reason I chose this solution is that it effectively patches the vulnerability without altering the library's behavior (i.e. what is saved in the database).

### DIFF
--- a/aiohttp_session/__init__.py
+++ b/aiohttp_session/__init__.py
@@ -18,13 +18,14 @@ class Session(MutableMapping):
     def __init__(self, identity, *, data, new, max_age=None):
         self._changed = False
         self._mapping = {}
-        self._identity = identity
+        self._identity = identity if data != {} else None
         self._new = new
+        self._new = new if data != {} else True
         self._max_age = max_age
         created = data.get('created', None) if data else None
         session_data = data.get('session', None) if data else None
 
-        if new or created is None:
+        if self._new or created is None:
             self._created = int(time.time())
         else:
             self._created = created
--- a/aiohttp_session/redis_storage.py
+++ b/aiohttp_session/redis_storage.py
@@ -47,7 +47,7 @@ class RedisStorage(AbstractStorage):
             with await self._redis as conn:
                 key = str(cookie)
                 data = await conn.get(self.cookie_name + '_' + key)
-                if data is None or data == b'{}':
+                if data is None:
                     return Session(None, data=None,
                                    new=True, max_age=self.max_age)
                 data = data.decode('utf-8')
--- a/aiohttp_session/redis_storage.py
+++ b/aiohttp_session/redis_storage.py
@@ -47,7 +47,7 @@ class RedisStorage(AbstractStorage):
             with await self._redis as conn:
                 key = str(cookie)
                 data = await conn.get(self.cookie_name + '_' + key)
-                if data is None:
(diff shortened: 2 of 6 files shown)


panagiks:
Running the scripts referenced in #272 will now produce the output:

```
First Login, Cookie:  Set-Cookie: FIXATED=e7f0169aabf141c5a142fed90a7b1424; Domain=127.0.0.1; HttpOnly; Path=/                                                                                                     
After Logout, Cookie:  Set-Cookie: FIXATED=""; Domain=127.0.0.1; expires=Thu, 01 Jan 1970 00:00:00 GMT; Max-Age=0; Path=/                                                                                          
Victim Login, Cookie:  Set-Cookie: FIXATED=ba72f10e160640bd9066a7d4882a2c8a; Domain=127.0.0.1; HttpOnly; Path=/
```

Effectively assigning a new cookie to the victim rather than re-registering the one presented.

asvetlov:
Yes, I see.
Looks like other storages are affected also.
Would you fix them all?

panagiks:
Ok I'm looking into it now :)

panagiks:
Ok, so I think the most safe way is to not rely on storages (as people may roll their own implementations for whatever reason) but better catch the issue at a lower level. In that though I made a couple changes in the Session class itself so it will now consider a session with data == {} to be new and will ignore any identity passed to it.

panagiks:
Ok, should be good now :) Let me know if it needs any change or modification.

asvetlov:
thanks!

PR #275, opened 2018-05-02
fix setup to read version on windows.

alexpantyukhin:
The same as for aiohttp-security in the commit 
https://github.com/aio-libs/aiohttp-security/commit/797d892ac123061dd92a70111e5aea7487973313


asvetlov:
thanks

PR #276, opened 2018-05-02
move checking runtime errors upper for redis_storage

alexpantyukhin:
I think that checking of runtime errors shoulb be at the start in ctor.


alexpantyukhin:
Actually the idea was that we don't need to do redundant actions (call `super().__init__`) when we raise `RuntimeError` . It really doesn't fix any bug, and doesn't improve performance much. I just thought that it would be correct :)

asvetlov:
Thank you for the contribution but I'm inclining to reject the issue.
Reasons are:
1. Parent constructor call is cheap
2. The current code is safe and correct
3. All other storages call base constructor at the very first line.

alexpantyukhin:
Oh, I see. Thank you for your comments and discussion ! :)

PR #278, opened 2018-05-07
CSRF support

bmwant:
## Rationale
Basic implementation of CSRF check as mentioned in https://github.com/aio-libs/aiohttp-session/issues/25#issuecomment-168665045.

@asvetlov let me know if we need to have more complex api (rather than just two coroutines) or additional checks e.g.
* checks via header `X_CSRF_Token`
* checking origin/referrer headers
* check same domain?
* check schema?


bmwant:
Ok, thanks. Will review source code of that library (did not know about that).

bmwant:
closing as redundant

PR #282, opened 2018-05-10
Implement new_session functionality

### LINKED ISSUE #281: [Feature] Requesting new session explicitly
## General

I think `aiohttp-session` should provide a way to explicitly request a new session, ignoring the existence of a cookie.

## Implementation

My initial thought is to pass an optional kwarg `new=True` to `get_session`, which in turn will pass it to the storage's `load_session` function to force it to return a new Session ignoring the presence of any cookie.

## Background

This is another window for Session Fixation, although this does not affect `aiohttp-session` itself like #272 but the application logic. I think that providing a way to explicitly request a new session (and encouraging the use of it in documentation and examples related to login functionality) would greatly benefit the users of `aiohttp-session` in terms of writing more secure code.

Looking into [OWASP's entry for Session Fixation](https://www.owasp.org/index.php/Session_fixation):

>  When authenticating a user, it [the vulnerable web application] doesn’t assign a new session ID, making it possible to use an existent session ID.

@asvetlov do you agree with adding this feature ? If so I can have a PR ready during the weekend (to also add documentation and usage examples).

panagiks:
Closes #281 .

Implements new_session functionality,

Update documentation to include new function.

Update demo code to use new_session for login view.

### DIFF
--- a/aiohttp_session/__init__.py
+++ b/aiohttp_session/__init__.py
@@ -120,6 +120,22 @@ async def get_session(request):
     return session
 
 
+def new_session(request):
+    storage = request.get(STORAGE_KEY)
+    if storage is None:
+        raise RuntimeError(
+            "Install aiohttp_session middleware "
+            "in your aiohttp.web.Application")
+    else:
+        session = storage.new_session()
+        if not isinstance(session, Session):
+            raise RuntimeError(
+                "Installed {!r} storage should return session instance "
+                "on .load_session() call, got {!r}.".format(storage, session))
+        request[SESSION_KEY] = session
+    return session
+
+
 def session_middleware(storage):
 
     if not isinstance(storage, AbstractStorage):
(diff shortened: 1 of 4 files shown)


panagiks:
I decided to make `new_session` a regular function (not a coroutine) since there is no reading/writing to the Storage at that point just the creation of an object and placement in the request dictionary.

If there's on objection with that implementation (i.e. you'd prefer to have it as a coroutine for uniformity with `get_session`) let me know :)

asvetlov:
I prefer a coroutine.
1. If we'll need a coroutine in a future it's hard to convert a public API
2. Making both `get_session()` and `new_session()` coroutines less confuses users.

panagiks:
Ok, I'll make the change (code and docs) in the evening :)

panagiks:
Should be good now. Let me know I any further changes are required :)

PR #295, opened 2018-06-11
Decode not nessesary here

### LINKED ISSUE #294: Decoding data is non nesessary and breakes binary coders
What reason for this line here?
https://github.com/aio-libs/aiohttp-session/blob/606cacafc47383f99cff39aba31f20bb5db3b879/aiohttp_session/redis_storage.py#L53

I want to store some bytes in session. Changed encoder  `    setup_session(app, RedisStorage(redis_pool, encoder=pickle.dumps, decoder=pickle.loads))`. And I have error on getting session:
```

    data = data.decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte

```
It also breakes `msgpack`

alex-eri:
fixes #294

### DIFF
--- a/aiohttp_session/redis_storage.py
+++ b/aiohttp_session/redis_storage.py
@@ -50,6 +50,8 @@ class RedisStorage(AbstractStorage):
                 if data is None:
                     return Session(None, data=None,
                                    new=True, max_age=self.max_age)
+                if self._decoder == json.loads:
+                    data = data.decode('utf-8')
                 try:
                     data = self._decoder(data)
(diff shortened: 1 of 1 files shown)


asvetlov:
The PR changes semantix.
Was "encoder/decoder are convertes between object and str".
Now "decoder is a converter from str to object if it is exactly json.loads, otherwise it converts from bytes to object. Encoder semantics is not changed".

Sorry,  cannot accept such change. It is at least not backward compatible.

PR #300, opened 2018-06-25
Fix the travis build badge and codecov badge in the docs

Mariatta:
Currently in http://aiohttp-session.readthedocs.io/en/latest/, the badges are saying "unknown".

With this fix, it will show the correct badges:

screen shot 2018-06-24 at 10 26 44 pm


asvetlov:
thanks

PR #374, opened 2019-06-21
[Doc] Return string not bytes for uniformity's sake

### LINKED ISSUE #368: document bug in homepage that always return a download
https://aiohttp-session.readthedocs.io/en/latest/
```
import time
from aiohttp import web
from aiohttp_session import get_session, setup
from aiohttp_session.cookie_storage import EncryptedCookieStorage

async def handler(request):
    session['last_visit'] = time.time()
    return web.Response(body=b'OK')

def init():
    app = web.Application()
    setup(app,
        EncryptedCookieStorage(b'Thirty  two  length  bytes  key.'))
    app.router.add_route('GET', '/', handler)
    return app

web.run_app(init())
```
`return web.Response(body=b'OK')` will return a download file to my chrome.

considering fix?

panagiks:
## What do these changes do?

In the main documentation page, change the return body of the 'Usage' example to pass a `string` to the `web.Response` object's constructor instead of `bytes` object.

This is not a real issue but since it has already confused one user (see related issue number section bellow) and it is the only place in `aiohttp-session` examples where `bytes` is provided to the `web.Response` constructor I suggest this change purely for the sake of uniformity (and to avoid any confusion).

None.

## Related issue number

Closes #368 .

## Checklist

- [x] Add a new news fragment into the `CHANGES` folder

That's only a docs change and the 'CHANGES' folder doesn't exist so the above are check as 'doesn't apply' ;)

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -37,7 +37,7 @@ A trivial usage example:
     async def handler(request):
         session = await get_session(request)
         session['last_visit'] = time.time()
-        return web.Response(body=b'OK')
+        return web.Response(body='OK')
 
     def init():
         app = web.Application()


asvetlov:
`body` **is** a binary.
Use `text` for strings please.

asvetlov:
Fixed by 2904c07

PR #404, opened 2019-11-12
Initial type hint implementation

### LINKED ISSUE #403: Add type hints
The issue is relatively easy.

Does anybody want to be a champion?

panagiks:
## What do these changes do?

Introduce type hints to `aio-libs/aiohttp-session`.

No behavioral, API (up to now, see notes bellow).

## Related issue number

Closes #403 

## Checklist

None apply.

## Why this is WiP

I've hit a few roadblocks here and would like a second opinion and took a few calls which I'd like to document and have ACKed :)

### Code change

I've changed 

```py
params = dict(self._cookie_params)
```

to 

```py
params: _TCookieParams = self._cookie_params.copy()
```

Reasoning: In the previous state, `mypy` recognized `params` as `Dict[Any, Any]` regardless of the fact that `self._cookie_params` was typed as `_TCookieParams`. Bonus: I think `.copy()` makes more clear what happens and why.

From my tests this shouldn't introduce any behavior change.

### Style change

I've changed the hanging indentation from first argument inline (right hanging) to first argument on new line (left hanging). Reasoning: on at least a couple occasions right hanging arguments were exceeding the 79th column with their type annotation and flake was complaining. Instead of changing only these instances I opted for consistency. 

### Things I've ignored

- imports that didn't have type annotations are ignored in mypy.ini
- kwargs to `aiohttp.StreamResponse.set_cookie` since the type for `httponly` & `secure` were fixed with aio-libs/aiohttp#4238
- tests: functions decorated with @pytest.fixture; for two reasons. The decorator is untyped, making the decorated functions also untyped. I took a few shots at writing a stub for the decorator with no success. Second, the actual fixture typing is not utilized anywhere, since its result is injected in the related tests and properly typed in the test function's definition.
- tests: conftest.py => ignored in it's entirety. Everything is a fixture.
- tests: Intentionally wrong functions. In at least a few test cases we create classes / functions that intentionally violate the expectations in order to test our error handling code. I have documented that the violation is intentional and ignored said functions.
- tests: `client.session.cookie_jar.update_cookies`. With the latest `aiohttp` version `update_cookies` reported it accepts `Union[Iterable[Tuple[str, BaseCookie[str]]], Mapping[str, BaseCookie[str]], BaseCookie[str]]` which does not include our very basic case of `Dict[str, str]` which comes from something like `{'AIOHTTP_SESSION': json.dumps(...)}`. From a look into `aiohttp` sources this could have been fixed with aio-libs/aiohttp#4252 so I've ignored these cases and documented that we should revisit them when the above is available.

### Errors still here (Second Opinion would help :) )

```sh
$ mypy aiohttp_session tests --strict
aiohttp_session/__init__.py:40: error: Missing type parameters for generic type "MutableMapping"
aiohttp_session/__init__.py:205: error: Exception must be derived from BaseException
aiohttp_session/nacl_storage.py:67: error: Argument 2 to "save_cookie" of "AbstractStorage" has incompatible type "Dict[Any, Any]"; expected "str"
Found 3 errors in 2 files (checked 20 source files)
```

- `Session` subclasses `MutableMapping` which is a generic type. I haven't found a way to define its type and didn't want to ignore it
- `session_middleware` catches `web.HTTPException` and assigns it to `response` thus duck typing it. This is done so that the session is saved in the related storage if it's changed regardless of whether the request raised an Exception. This causes 'raise response' to fail as `response` is typed as `web.StreamResponse`.
- `nacl_storage` wrong argument to `save_cookie`. This only worked as a side-effect.

`nacl_storage` does on `save_session`:

```py
if session.empty:
            return self.save_cookie(response, session._mapping,
                                    max_age=session.max_age)
```

and `AbstractStorage`'s `save_cookie` checks:

```py
if not cookie_data:
        response.del_cookie(
                self._cookie_name,
                domain=params["domain"],
                path=params["path"],
                )
```

But here `nacl_storage` technically is wrong and should IMO just pass `''`.

I realize the above are more than a lot so, take your time :) Let me know for any change required.


panagiks:
Python 3.5 seems unhappy with inline type annotations: `x: str = ""`.

I'll have a look later in the day.

aiohttp_session/__init__.py
```diff
@@ -8,7 +8,34 @@
...
+    Callable,
+    Dict,
+    Hashable,
+    Iterator,
+    Optional,
+    Tuple,
+)
+
+if TYPE_CHECKING:  # pragma: no cover
+    from mypy_extensions import TypedDict
```
asvetlov: Please use `typing_extensions` library instead.
Also, it doesn't need to be guarded by `TYPE_CHECKING` IMHO.

aiohttp_session/__init__.py
```diff
@@ -175,41 +211,50 @@ def session_middleware(storage):
...
-                 encoder=json.dumps, decoder=json.loads):
+    def __init__(
+        self, *,
+        cookie_name: str = "AIOHTTP_SESSION",
+        domain: Optional[str] = None,
+        max_age: Optional[int] = None,
+        path: str = '/',
+        secure: Optional[bool] = None,
+        httponly: bool = True,
+        encoder: Callable[..., str] = json.dumps,
```
asvetlov:
```suggestion
        encoder: Callable[[Dict[str, Any]], str] = json.dumps,
```
Encoder is called with JSON only in the library

aiohttp_session/__init__.py
```diff
@@ -175,41 +211,50 @@ def session_middleware(storage):
...
+        decoder: Callable[..., Dict[Any, Any]] = json.loads
```
asvetlov:
```suggestion
        decoder: Callable[str, Dict[str, Any]] = json.loads
```

IFAIK the session keys are always strings

aiohttp_session/__init__.py
```diff
@@ -256,23 +311,35 @@ class SimpleCookieStorage(AbstractStorage):
...
-                 encoder=json.dumps, decoder=json.loads):
+    def __init__(
+        self, *,
+        cookie_name: str = "AIOHTTP_SESSION",
+        domain: Optional[str] = None,
+        max_age: Optional[int] = None,
+        path: str = '/',
+        secure: Optional[bool] = None,
+        httponly: bool = True,
+        encoder: Callable[..., str] = json.dumps,
```
asvetlov: The same as above

aiohttp_session/__init__.py
```diff
@@ -256,23 +311,35 @@ class SimpleCookieStorage(AbstractStorage):
...
+        decoder: Callable[..., Dict[Any, Any]] = json.loads
```
asvetlov: Please avoid the ellipsis

aiohttp_session/cookie_storage.py
```diff
@@ -12,10 +15,18 @@ class EncryptedCookieStorage(AbstractStorage):
...
+    def __init__(
+        self,
+        secret_key: Union[str, bytes, bytearray], *,
+        cookie_name: str = "AIOHTTP_SESSION",
+        domain: Optional[str] = None,
+        max_age: Optional[int] = None,
+        path: str = '/',
+        secure: Optional[bool] = None,
+        httponly: bool = True,
+        encoder: Callable[..., str] = json.dumps,
```
asvetlov: Ellipsis

aiohttp_session/cookie_storage.py
```diff
@@ -12,10 +15,18 @@ class EncryptedCookieStorage(AbstractStorage):
...
+        decoder: Callable[..., Dict[Any, Any]] = json.loads
```
asvetlov: Ellipsis

aiohttp_session/memcached_storage.py
```diff
@@ -1,25 +1,40 @@
...
+        memcached_conn: 'aiomcache.Client', *,
+        cookie_name: str = "AIOHTTP_SESSION",
+        domain: Optional[str] = None,
+        max_age: Optional[int] = None,
+        path: str = '/',
+        secure: Optional[bool] = None,
+        httponly: bool = True,
+        key_factory: Callable[[], str] = lambda: uuid.uuid4().hex,
+        encoder: Callable[..., str] = json.dumps,
+        decoder: Callable[..., Dict[Any, Any]] = json.loads
```
asvetlov: Ellipsis

aiohttp_session/nacl_storage.py
```diff
@@ -14,21 +17,29 @@ class NaClCookieStorage(AbstractStorage):
...
+        self,
+        secret_key: bytes, *,
+        cookie_name: str = "AIOHTTP_SESSION",
+        domain: Optional[str] = None,
+        max_age: Optional[int] = None,
+        path: str = '/',
+        secure: Optional[bool] = None,
+        httponly: bool = True,
+        encoder: Callable[..., str] = json.dumps,
+        decoder: Callable[..., Dict[Any, Any]] = json.loads
```
asvetlov: Ellipsis

aiohttp_session/redis_storage.py
```diff
@@ -7,17 +7,29 @@
...
+        redis_pool: 'aioredis.commands.Redis', *,
+        cookie_name: str = "AIOHTTP_SESSION",
+        domain: Optional[str] = None,
+        max_age: Optional[int] = None,
+        path: str = '/',
+        secure: Optional[bool] = None,
+        httponly: bool = True,
+        key_factory: Callable[[], str] = lambda: uuid.uuid4().hex,
+        encoder: Callable[..., str] = json.dumps,
+        decoder: Callable[..., Dict[Any, Any]] = json.loads
```
asvetlov: Ellipsis

mypy.ini
```diff
@@ -0,0 +1,29 @@
+[mypy-nacl]
```
asvetlov: Please merge `mypy.ini` and `pytest.ini` into `setup.cfg`.
The single config file is better than two :)

aiohttp_session/__init__.py
panagiks: Hadn't seen the `typing_extensions` lib up to now. With this it makes sense to not need the `TYPE_CHECKING` guard :)

aiohttp_session/__init__.py
panagiks: Sorry for the ellipsis spam, they were a "revisit this later" for myself ... which I forgot to revisit :s All should be cleaned up now.

mypy.ini
panagiks: Better indeed :)

aiohttp_session/__init__.py
panagiks: Fixed :)

aiohttp_session/__init__.py
panagiks: Fixed :)

aiohttp_session/__init__.py
panagiks: Fixed :)

panagiks:
Pipeline failure is expected after adding mypy check to travis (see "Errors still here" section for the last 6 remaining errors).

asvetlov:
Have you a chance to fix mypy errors shown by Travis?

panagiks:
@asvetlov should be good now :) (Sorry for the delay didn't have time to check it the past couple days). Only please have a look at my last commit (change in NaCl storage) and let me know if you are OK with the change. (Rationale for the change is available in the initial description of the PR)

aiohttp_session/__init__.py
```diff
@@ -5,21 +5,50 @@
...
+    def __init__(
+        self,
+        identity: Optional[Any], *,
+        data: Optional[Dict[Any, Any]],
+        new: bool,
+        max_age: Optional[int] = None
+    ) -> None:
         self._changed = False
-        self._mapping = {}
+        self._mapping = {}  # type: Dict[Any, Any]
```
Dreamsorcerer, line 51: We can inline now, as we no longer support 3.5.

Dreamsorcerer:
I realise this is quite old, are you interested in finishing it off and getting it merged?

panagiks:
Hello :) Sure, I'll have a look on the left open points in the next 1-2 weeks. IIRC there were a couple of open points on whether somethings should be ignored or not, I'll make sure to refresh / rephrase those questions; just a heads up that this may require some input / opinion.

Dreamsorcerer:
Great, I'll go through and add some nitpicks, but first thing would be to merge master in order to get the tests running again.

setup.cfg
```diff
@@ -0,0 +1,49 @@
+[tool:pytest]
+filterwarnings=
+    error
+
+[mypy]
```
Dreamsorcerer, line 5: I'd be tempted to keep this separate in `.mypy.ini` as it's a substantial amount of config.

setup.py
```diff
@@ -14,7 +14,7 @@
 def read(f):
     return open(os.path.join(os.path.dirname(__file__), f)).read().strip()
 
-install_requires = ['aiohttp>=3.0.1']
+install_requires = ['aiohttp>=3.0.1', 'typing_extensions>=3.6.5']
```
Dreamsorcerer, line 17: Only in Python <3.8.
See: https://github.com/aio-libs/aiohttp/blob/3.8/setup.py#L73

tests/conftest.py
```diff
@@ -1,3 +1,4 @@
+# type: ignore
```
Dreamsorcerer, line 1: Ignore what and why?
Please use the `# type: ignore[code]` format and include a comment explaining why (if it is not obvious).

tests/test_abstract_storage.py
```diff
@@ -3,31 +3,44 @@
...
-def make_cookie(client, data):
+def make_cookie(client: TestClient, data: Dict[Any, Any]) -> None:
     session_data = {
         'session': data,
         'created': int(time.time())
     }
 
     value = json.dumps(session_data)
-    client.session.cookie_jar.update_cookies({'AIOHTTP_SESSION': value})
+    # Ignoring type until aiohttp#4252 is released
```
Dreamsorcerer, line 24: Can be removed.

tests/test_abstract_storage.py
```diff
@@ -3,31 +3,44 @@
...
-async def test_max_age_also_returns_expires(aiohttp_client):
+async def test_max_age_also_returns_expires(
+    aiohttp_client: _TAiohttpClient
+) -> None:
 
-    async def handler(request):
-        time.time.return_value = 0.0
+    async def handler(request: web.Request) -> web.StreamResponse:
+        # Ignoring type since time.time is mocked in this context
+        time.time.return_value = 0.0  # type: ignore
```
Dreamsorcerer, line 43: Maybe a task for another PR, but ideally we shouldn't be referencing mock objects in this way, we should have a direct reference to the mock which we can safely modify without risk of affecting it's state.

tests/test_cookie_storage.py
```diff
@@ -2,30 +2,40 @@
...
-def make_cookie(client, data):
+def make_cookie(client: TestClient, data: Dict[Any, Any]) -> None:
     session_data = {
         'session': data,
         'created': int(time.time())
     }
 
     value = json.dumps(session_data)
-    client.session.cookie_jar.update_cookies({'AIOHTTP_SESSION': value})
+    # Ignoring type until aiohttp#4252 is released
```
Dreamsorcerer, line 23: Can be removed.

tests/test_encrypted_cookie_storage.py
```diff
@@ -5,69 +5,94 @@
...
+    data: Dict[Any, Any]
+) -> None:
     session_data = {
         'session': data,
         'created': int(time.time())
     }
 
     cookie_data = json.dumps(session_data).encode('utf-8')
-    data = fernet.encrypt(cookie_data).decode('utf-8')
+    encrypted_data = fernet.encrypt(cookie_data).decode('utf-8')  # type: str
```
Dreamsorcerer, line 36: Inline, but should be inferred (unless `cryptography` is not typed).

tests/test_encrypted_cookie_storage.py
```diff
@@ -5,69 +5,94 @@
...
 
-    client.session.cookie_jar.update_cookies({'AIOHTTP_SESSION': data})
+    # Ignoring type until aiohttp#4252 is released
```
Dreamsorcerer, line 38: Can be removed.

tests/test_encrypted_cookie_storage.py
```diff
@@ -5,69 +5,94 @@
...
-def decrypt(fernet, cookie_value):
+def decrypt(fernet: Fernet, cookie_value: str) -> Any:
     assert type(cookie_value) == str
     return json.loads(
         fernet.decrypt(cookie_value.encode('utf-8')).decode('utf-8')
     )
 
 
+# pytest.fixture decorator strips the typing of the decorated function
+@no_type_check
```
Dreamsorcerer, line 62: Shouldn't be using this. This is used to flag that annotations are not meant to be type definitions.

I see there is typing support in pytest, so maybe this has already been fixed by them. Otherwise, let's figure out another solution.

tests/test_encrypted_cookie_storage.py
```diff
@@ -5,69 +5,94 @@
...
 @pytest.fixture
-def fernet_and_key():
+def fernet_and_key() -> Tuple[Fernet, bytes]:
     key = Fernet.generate_key()
     fernet = Fernet(key)
     return fernet, base64.urlsafe_b64decode(key)
 
 
+# pytest.fixture decorator strips the typing of the decorated function
+@no_type_check
```
Dreamsorcerer, line 71: Remove.

tests/test_encrypted_cookie_storage.py
```diff
@@ -5,69 +5,94 @@
...
 @pytest.fixture
-def fernet(fernet_and_key):
+def fernet(fernet_and_key: Tuple[Fernet, bytes]) -> Fernet:
     return fernet_and_key[0]
 
 
+# pytest.fixture decorator strips the typing of the decorated function
+@no_type_check
```
Dreamsorcerer, line 78: Remove.

tests/test_encrypted_cookie_storage.py
```diff
@@ -162,18 +203,25 @@ async def logout(request):
     evil_cookie = resp.cookies['AIOHTTP_SESSION'].value
     resp = await client.delete('/')
     assert resp.cookies['AIOHTTP_SESSION'].value == ""
-    client.session.cookie_jar.update_cookies({'AIOHTTP_SESSION': evil_cookie})
+    # Ignoring type until aiohttp#4252 is released
```
Dreamsorcerer, line 206: Remove.

tests/test_encrypted_cookie_storage.py
```diff
@@ -193,7 +241,10 @@ async def handler(request):
     assert 'AIOHTTP_SESSION' in resp.cookies
     cookie = resp.cookies['AIOHTTP_SESSION'].value
     await asyncio.sleep(MAX_AGE + 1)
-    client.session.cookie_jar.update_cookies({'AIOHTTP_SESSION': cookie})
+    # Ignoring type until aiohttp#4252 is released
```
Dreamsorcerer, line 244: Remove.

tests/test_get_session.py
```diff
@@ -35,15 +38,22 @@ async def load_session(self, request):
...
 
-async def test_get_new_session():
+async def test_get_new_session() -> None:
     req = make_mocked_request('GET', '/')
     session = Session('identity', data=None, new=False)
 
     class Storage(AbstractStorage):
-        async def load_session(self, request):
+        # Ignoring typing since return type is on purpose wrong
+        @no_type_check
```
Dreamsorcerer, line 47: Use a `type: ignore[code]` instead for same reasons above.
We could probably also just change `pass` to `return {}` or similar and avoid the errors altogether.

tests/test_get_session.py
```diff
@@ -62,17 +72,26 @@ async def test_get_new_session_no_storage():
...
     class Storage(AbstractStorage):
-        async def new_session(self):
+        # Ignoring typing since return type is on purpose wrong
+        @no_type_check
+        async def new_session(self) -> str:
             return ''
 
-        async def load_session(self, request):
+        # Ignoring typing since return type is on purpose wrong
+        @no_type_check
```
Dreamsorcerer, line 85: Same for these 2.

tests/test_memcached_storage.py
```diff
@@ -26,17 +42,29 @@ async def make_cookie(client, memcached, data):
     key = uuid.uuid4().hex
     storage_key = ('AIOHTTP_SESSION_' + key).encode('utf-8')
     await memcached.set(storage_key, bytes(value, 'utf-8'))
-    client.session.cookie_jar.update_cookies({'AIOHTTP_SESSION': key})
+    # Ignoring type until aiohttp#4252 is released
```
Dreamsorcerer, line 45: Remove.

tests/test_memcached_storage.py
```diff
@@ -26,17 +42,29 @@ async def make_cookie(client, memcached, data):
...
-async def make_cookie_with_bad_value(client, memcached):
+async def make_cookie_with_bad_value(
+    client: TestClient,
+    memcached: aiomcache.Client
+) -> None:
     key = uuid.uuid4().hex
     storage_key = ('AIOHTTP_SESSION_' + key).encode('utf-8')
     await memcached.set(storage_key, b'')
-    client.session.cookie_jar.update_cookies({'AIOHTTP_SESSION': key})
+    # Ignoring type until aiohttp#4252 is released
```
Dreamsorcerer, line 58: Remove

tests/test_memcached_storage.py
```diff
@@ -168,30 +213,36 @@ async def handler(request):
...
+) -> None:
 
-    async def handler(request):
+    async def handler(request: web.Request) -> web.StreamResponse:
         session = await get_session(request)
         assert session.new
         return web.Response(body=b'OK')
 
     client = await aiohttp_client(create_app(handler, memcached))
+    # Ignoring type until aiohttp#4252 is released
```
Dreamsorcerer, line 226: Remove

tests/test_memcached_storage.py
```diff
@@ -225,14 +280,20 @@ async def logout(request):
     evil_cookie = resp.cookies['AIOHTTP_SESSION'].value
     resp = await client.delete('/')
     assert resp.cookies['AIOHTTP_SESSION'].value == ""
-    client.session.cookie_jar.update_cookies({'AIOHTTP_SESSION': evil_cookie})
+    # Ignoring type until aiohttp#4252 is released
```
Dreamsorcerer, line 283: Remove

tests/test_nacl_storage.py
```diff
@@ -29,37 +40,52 @@ def make_cookie(client, secretbox, data):
     data = secretbox.encrypt(cookie_data, nonce,
                              encoder=Base64Encoder).decode('utf-8')
 
-    client.session.cookie_jar.update_cookies({'AIOHTTP_SESSION': data})
+    # Ignoring type until aiohttp#4252 is released
```
Dreamsorcerer, line 43: Remove

tests/test_nacl_storage.py
```diff
@@ -29,37 +40,52 @@ def make_cookie(client, secretbox, data):
...
+def decrypt(secretbox: nacl.secret.SecretBox, cookie_value: str) -> Any:
     assert type(cookie_value) == str
     return json.loads(
         secretbox.decrypt(cookie_value.encode('utf-8'),
                           encoder=Base64Encoder).decode('utf-8')
     )
 
 
+# pytest.fixture decorator strips the typing of the decorated function
+@no_type_check
```
Dreamsorcerer, line 69: Remove

tests/test_nacl_storage.py
```diff
@@ -29,37 +40,52 @@ def make_cookie(client, secretbox, data):
...
 @pytest.fixture
-def secretbox(key):
+def secretbox(key: bytes) -> nacl.secret.SecretBox:
     return nacl.secret.SecretBox(key)
 
 
+# pytest.fixture decorator strips the typing of the decorated function
+@no_type_check
```
Dreamsorcerer, line 76: Remove

tests/test_nacl_storage.py
```diff
@@ -152,14 +194,20 @@ async def logout(request):
     evil_cookie = resp.cookies['AIOHTTP_SESSION'].value
     resp = await client.delete('/')
     assert resp.cookies['AIOHTTP_SESSION'].value == ""
-    client.session.cookie_jar.update_cookies({'AIOHTTP_SESSION': evil_cookie})
+    # Ignoring type until aiohttp#4252 is released
```
Dreamsorcerer, line 197: Remove

tests/test_nacl_storage.py
```diff
@@ -182,24 +230,33 @@ async def handler(request):
...
+    async def handler(request: web.Request) -> web.StreamResponse:
         session = await get_session(request)
         assert isinstance(session, Session)
         assert session.new
         assert {} == session
         return web.Response(body=b'OK')
 
     client = await aiohttp_client(create_app(handler, key))
-    client.session.cookie_jar.update_cookies({'AIOHTTP_SESSION': 'bad key'})
+    # Ignoring type until aiohttp#4252 is released
```
Dreamsorcerer, line 246: Remove

tests/test_nacl_storage.py
```diff
@@ -239,7 +299,10 @@ async def handler(request):
     assert 'AIOHTTP_SESSION' in resp.cookies
     cookie = resp.cookies['AIOHTTP_SESSION'].value
     await asyncio.sleep(MAX_AGE + 1)
-    client.session.cookie_jar.update_cookies({'AIOHTTP_SESSION': cookie})
+    # Ignoring type until aiohttp#4252 is released
```
Dreamsorcerer, line 302: Remove

tests/test_path_domain.py
```diff
@@ -3,24 +3,40 @@
...
+    data: Any,
+    path: Optional[str] = None,
+    domain: Optional[str] = None
+) -> None:
     session_data = {
         'session': data,
         'created': int(time.time())
     }
-    C = cookies.SimpleCookie()
+    C = cookies.SimpleCookie()  # type: cookies.SimpleCookie[str]
```
Dreamsorcerer, line 27: Inline

tests/test_redis_storage.py
```diff
@@ -28,17 +44,29 @@ async def make_cookie(client, redis, data):
     key = uuid.uuid4().hex
     with await redis as conn:
         await conn.set('AIOHTTP_SESSION_' + key, value)
-    client.session.cookie_jar.update_cookies({'AIOHTTP_SESSION': key})
+    # Ignoring type until aiohttp#4252 is released
```
Dreamsorcerer, line 47: Remove

tests/test_redis_storage.py
```diff
@@ -28,17 +44,29 @@ async def make_cookie(client, redis, data):
...
-async def make_cookie_with_bad_value(client, redis):
+async def make_cookie_with_bad_value(
+    client: TestClient,
+    redis: aioredis.commands.Redis
+) -> None:
     key = uuid.uuid4().hex
     with await redis as conn:
         await conn.set('AIOHTTP_SESSION_' + key, '')
-    client.session.cookie_jar.update_cookies({'AIOHTTP_SESSION': key})
+    # Ignoring type until aiohttp#4252 is released
```
Dreamsorcerer, line 60: Remove

tests/test_redis_storage.py
```diff
@@ -210,29 +262,36 @@ async def handler(request):
...
+) -> None:
 
-    async def handler(request):
+    async def handler(request: web.Request) -> web.StreamResponse:
         session = await get_session(request)
         assert session.new
         return web.Response(body=b'OK')
 
     client = await aiohttp_client(create_app(handler, redis))
+    # Ignoring type until aiohttp#4252 is released
```
Dreamsorcerer, line 275: Remove

tests/test_redis_storage.py
```diff
@@ -265,57 +328,65 @@ async def logout(request):
     evil_cookie = resp.cookies['AIOHTTP_SESSION'].value
     resp = await client.delete('/')
     assert resp.cookies['AIOHTTP_SESSION'].value == ""
-    client.session.cookie_jar.update_cookies({'AIOHTTP_SESSION': evil_cookie})
+    # Ignoring type until aiohttp#4252 is released
```
Dreamsorcerer, line 331: Remove

tests/test_response_types.py
```diff
@@ -28,9 +34,11 @@ async def stream_response(request):
     assert SESSION_KEY.upper() not in resp.cookies
 
 
-async def test_bad_response_type(aiohttp_client):
+async def test_bad_response_type(aiohttp_client: _TAiohttpClient) -> None:
 
-    async def bad_response(request):
+    # Ignoring typing since return type is on purpose wrong
+    @no_type_check
```
Dreamsorcerer, line 40: Remove

tests/test_session_middleware.py
```diff
@@ -3,6 +3,7 @@
 from aiohttp_session import session_middleware
 
 
-async def test_session_middleware_bad_storage():
+async def test_session_middleware_bad_storage() -> None:
     with pytest.raises(RuntimeError):
-        session_middleware(None)
+        # Ignoring typing since return type is on purpose wrong
+        session_middleware(None)  # type: ignore
```
Dreamsorcerer, line 9: Should probably read `Parameter type is wrong on purpose`.
Also, include the code: `# type: ignore[code]`.

tests/typedefs.py
```diff
@@ -0,0 +1,6 @@
+from aiohttp import web
+from aiohttp.test_utils import TestClient
+
+from typing import Awaitable, Callable
+
+_TAiohttpClient = Callable[[web.Application], Awaitable[TestClient]]
```
Dreamsorcerer, line 6: Isn't this for `aiohttp.pytest_plugin`? Would probably be better if we included the type in `aiohttp` directly and then import it into other projects, rather than needing to redefine it in every project.

Dreamsorcerer (changes requested):
Looking good, these comments should cover all the small changes that are needed.

panagiks:
Thank you for the feedback :)

> Great, I'll go through and add some nitpicks, but first thing would be to merge master in order to get the tests running again.

On it, any objection to rebasing on top of master instead (personal preference, it makes the history a bit more 'clear' to me)?

aiohttp_session/nacl_storage.py
```diff
@@ -46,9 +57,14 @@ async def load_session(self, request):
...
-    async def save_session(self, request, response, session):
+    async def save_session(
+        self,
+        request: web.Request,
+        response: web.StreamResponse,
+        session: Session
+    ) -> None:
         if session.empty:
-            return self.save_cookie(response, session._mapping,
+            return self.save_cookie(response, '',
```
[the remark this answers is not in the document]
panagiks: The reasoning is explained in the last bullet point of the MR description, sort version:

the `save_cookie` interface expects a string as the second argument see [here](https://github.com/aio-libs/aiohttp-session/blob/1127e80c039a876398ec18fc2c5d448229cf8214/aiohttp_session/cookie_storage.py#L51) for example.

The old code  worked as a side-effect, because `save_cookie` checks if `cookie_data` (the second argument) is 'Truthy' before using it (see [here](https://github.com/aio-libs/aiohttp-session/blob/1127e80c039a876398ec18fc2c5d448229cf8214/aiohttp_session/__init__.py#L243)) and if `session.empty` is `True` then `session._mapping` is an empty dictionary, which is not `Truthy` thus achieving the same result as an empty string, but violating the `save_cookie` type interface.

tests/test_abstract_storage.py
panagiks: I would keep it for another PR too, I've tried to keep code changes as minimal as possible in this one.

tests/typedefs.py
panagiks: > Isn't this for aiohttp.pytest_plugin?

Yes.

> Would probably be better if we included the type in aiohttp directly and then import it into other projects, rather than needing to redefine it in every project.

True, but given the current wait for `aiohttp 3.8` I would opt to add it now, with a note to replace it when `aiohttp` provides typing for that. (will also check whether typing for this is already provided in `aiohttp`, since it has been a while :D )

panagiks:
RE: changed for aio-libs:aiohttp#4252

The changes were merged after the last `3.6.x` release of `aiohttp` (`3.6.2` released on Oct 9 2020 and changes were merged on Oct 26 2020) and we have `aiohttp` pinned to `<3.7` (see [here](https://github.com/aio-libs/aiohttp-session/blob/1127e80c039a876398ec18fc2c5d448229cf8214/requirements-dev.txt#L15)) so I don't think the points can be removed until `3.8.x` is out & `aiohttp-session` changes the required version.

requirements-dev.txt
```diff
@@ -17,3 +17,5 @@ multidict==4.5.2
 chardet==3.0.4
 yarl==1.3.0
 aiomcache==0.6.0
+mypy==0.740
```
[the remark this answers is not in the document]
panagiks: Latest `mypy` version introduces some new errors that seem to come from implicit equality of `Session` (as a `MutableMapping`) to `dict`, for example in the tests there are a lot of `session == {}` assertions which with the latest `mypy` throw a `comparison-overlap` error (comparing `Session` to `Dict[x,x]`).

I'll look into fixing them (this may take a bit more :s ).

Dreamsorcerer:
> On it, any objection to rebasing on top of master instead (personal preference, it makes the history a bit more 'clear' to me)?

Not really, however you like. It'll end up squash committed when the PR is merged, so it won't make any difference to the history.

Dreamsorcerer:
> RE: changes for aio-libs:aiohttp#4252
> 
> The changes were merged after the last `3.6.x` release of `aiohttp` (`3.6.2` released on Oct 9 2020 and changes were merged on Oct 26 2020) and we have `aiohttp` pinned to `<3.7` (see [here](https://github.com/aio-libs/aiohttp-session/blob/1127e80c039a876398ec18fc2c5d448229cf8214/requirements-dev.txt#L15)) so I don't think the points can be removed until `3.8.x` is out & `aiohttp-session` changes the required version.

Ah, good point, I didn't check the version it was merged into.

tests/typedefs.py
Dreamsorcerer: I couldn't see anything, but feel free to make an extra PR to add it.

Agreed that we'll need this here as well for now.

aiohttp_session/__init__.py
```diff
@@ -168,48 +200,57 @@ async def factory(request, handler):
         if session is not None:
             if session._changed:
                 await storage.save_session(request, response, session)
-        if raise_response:
+        if isinstance(response, web.HTTPException):
```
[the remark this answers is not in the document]
panagiks: `raise_response` is only set to `True` in the try-except block above, the only 'class' of exceptions caught there is `web.HTTPException` see [here](https://github.com/aio-libs/aiohttp-session/blob/1127e80c039a876398ec18fc2c5d448229cf8214/aiohttp_session/__init__.py#L154) so I think this is 1:1 as far as logic is concerned.

tests/conftest.py
panagiks: For the why part (from the initial PR description, section 'Things I've ignored'):

```
- tests: functions decorated with @pytest.fixture; for two reasons. The decorator is untyped, making the decorated functions also untyped. I took a few shots at writing a stub for the decorator with no success. Second, the actual fixture typing is not utilized anywhere, since its result is injected in the related tests and properly typed in the test function's definition.
- tests: conftest.py => ignored in it's entirety. Everything is a fixture.
```

But I think I saw some on some of your notes something for the fixtures so I will revisit my approach on that (and make sure I don't leave any other 'bare' type ignores)

aiohttp_session/__init__.py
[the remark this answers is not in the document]
Dreamsorcerer: No, the handler can return `HTTPException` at `response = await handler(request)`.
You'll notice that this is a subclass of `Response` in the 3.x series.

While we encourage people to `raise` an exception instead of returning it, we should ensure backwards compatibility until we move to 4.x.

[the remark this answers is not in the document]
Dreamsorcerer: So, if `response = await handler(request)` returns a `HTTPException`, the new code would `raise` it, while the old code would `return` it.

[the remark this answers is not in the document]
panagiks: Ah, I see. I focused on the wrong part of the change. Thank you :)

aiohttp_session/__init__.py
```diff
@@ -219,23 +260,33 @@ def _get_session_data(self, session):
...
+        request: web.Request,
+        response: web.StreamResponse,
+        session: Session
+    ) -> None:
         pass
 
-    def load_cookie(self, request):
-        cookie = request.cookies.get(self._cookie_name)
+    def load_cookie(self, request: web.Request) -> Optional[str]:
+        cookie = request.cookies.get(self._cookie_name)  # type: Optional[str]
```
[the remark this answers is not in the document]
panagiks: Tested again, was not inferred.

aiohttp_session/__init__.py
```diff
@@ -219,23 +260,33 @@ def _get_session_data(self, session):
...
 
-    def save_cookie(self, response, cookie_data, *, max_age=None):
-        params = dict(self._cookie_params)
+    def save_cookie(
+        self,
+        response: web.StreamResponse,
+        cookie_data: str, *,
+        max_age: Optional[int] = None
+    ) -> None:
+        params = self._cookie_params.copy()  # type: _TCookieParams
```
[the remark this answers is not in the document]
panagiks: Tested again, was indeed inferred & I removed the annotation.

requirements-dev.txt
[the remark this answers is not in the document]
panagiks: Opted to `cast`ing the dict side of the comparisons to the appropriate `MutableMapping`, if you prefer ignoring them or centralizing the definition let me know :)

tests/test_encrypted_cookie_storage.py
panagiks: Was indeed inferred, removed the annotation :)

aiohttp_session/__init__.py
```diff
@@ -168,48 +200,57 @@ async def factory(request, handler):
...
-                                   path=path,
-                                   secure=secure,
-                                   httponly=httponly)
+        self._cookie_params = dict(
+            domain=domain,
+            max_age=max_age,
+            path=path,
+            secure=secure,
+            httponly=httponly
+        )  # type: _TCookieParams
```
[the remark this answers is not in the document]
panagiks: lgtm doesn't seem to like the `_TCookieParams` way :s (see alert about calling non-callable) will revert back to `{}` & annotation.

aiohttp_session/__init__.py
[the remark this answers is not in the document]
Dreamsorcerer: OK, I see what's happening, the @reify decorator on the cookies attribute is losing the info.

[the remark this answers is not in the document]
Dreamsorcerer: OK, looks like it's already fixed in 3.7, so let's add a `# TODO` so we can drop the annotation later.

aiohttp_session/__init__.py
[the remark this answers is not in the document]
Dreamsorcerer: I saw that, but it appears to be wrong, I'd just ignore the "error". As long as we have a test case that runs this code, we can be sure that it's working.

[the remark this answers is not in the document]
Dreamsorcerer: https://github.com/github/codeql/issues/5470

panagiks:
Recap of open points (will probably get some time mid/end of the week to resume):

- aiohttp#4238 (will let as-is and update when aiohttp 3.8 is out)
- `pytest.fixture`s (I will have to do a refresh on pytest typing, I don't remember a thing from the last time I touched this :D )
- mypy errors
  - `no-any-unimported` => `aioredis` and `aiomcache` are untyped which causes these errors, I will look a bit more into it but probably I'll add one 'broad' ignore for it in the tests and targeted ignores in the actual code (if there is no objection)
  - Fernet => `cryptography.fernet.Fernet` seems to expect `bytes` (based on its declared typing at least) but we pass it `str|bytes` (see [here](https://github.com/aio-libs/aiohttp-session/blob/1127e80c039a876398ec18fc2c5d448229cf8214/aiohttp_session/cookie_storage.py#L24-L28)), I will probably `.encode()` in case of string input.

Thank you for the feedback and help up to now :)

Dreamsorcerer:
>       * `no-any-unimported` => `aioredis` and `aiomcache` are untyped which causes these errors, I will look a bit more into it but probably I'll add one 'broad' ignore for it in the tests and targeted ignores in the actual code (if there is no objection)

There's only 3 errors in the main project, so just ignore those individually. Then change the strictness setting for tests in the config.

e.g. Add under the default block:
```
[mypy-tests.*]
disallow_any_unimported = False
```

Dreamsorcerer:
OK, I've looked through everything again. Happy to merge after the last couple of things are sorted. I'll probably make some followup PRs just to tighten up some typing afterwards.

panagiks:
Alright that should be it with the requested changes :)

Notes:
* Ignoring `disallow-any-unimported` on tests didn't work when the option was set to true in general config so I went the other way around and enabled `disallow-any-unimported` only on library code
* Some pytest fixtures accepted a `request` argument which was not used anywhere in their code, was not imported as a fixture from anywhere and there was no indication of what it was but seemed as a leftover. I dropped it.
* Codecov fails on the `if sys.version_info >= (3, 8):` check, should I add an ignore for it or do we just let it as is since it will be covered when we add testing on python 3.8 ?

Dreamsorcerer:
>     * Ignoring `disallow-any-unimported` on tests didn't work when the option was set to true in general config so I went the other way around and enabled `disallow-any-unimported` only on library code

Probably needs some `__init__.py`s, I'll try it out in a followup PR.

>     * Codecov fails on the `if sys.version_info >= (3, 8):` check, should I add an ignore for it or do we just let it as is since it will be covered when we add testing on python 3.8 ?

@webknjaz Can you disable/tweak the codecov requirement, or force merge the PR please?
The missing parts are due to some version checks being added, and I've not got the tests running for Python 3.8+ yet (something I'll look at after this).

PR #448, opened 2020-04-13
Allow passing Fernet to Encrypted Cookie Storage

### LINKED ISSUE #447: b64_encode needed in README?
The README takes the output of `Fernet.generate_key()` and then passes it through `urlsafe_b64decode()`. But, the cryptography documentation says "A URL-safe base64-encoded 32-byte key." is needed for `Fernet`, and then passes the result from `generate_key()` without encoding it again.
https://cryptography.io/en/latest/fernet/#cryptography.fernet.Fernet

So, surely the result of `generate_key()` is already in the correct format, and the `urlsafe_b64decode()` is just redundant.

~~The result of `generate_key()` is already 32 bytes, using 43 characters. With the README code, the result ends up at 59 characters (which is clearly more than 32 bytes b64 encoded).~~

Dreamsorcerer:
Allows passing a complete Fernet to constructor to avoid any guesswork around b64encoding.

Fixes #447


asvetlov (approved):
LGTM

PR #449, opened 2020-04-13
Add samesite parameter for cookies.

Dreamsorcerer:
Adds support for samesite.


HakuG:
Getting the problem with Chrome now that this isn't being set. Can we merge this?

Dreamsorcerer:
Seems to depend on a new release of aiohttp happening. So, I'd followup over there first:
https://github.com/aio-libs/aiohttp/issues/4516

NotSqrt:
Hello !

aiohttp v3.7 has now been released, with `SameSite` support : https://github.com/aio-libs/aiohttp/blob/master/CHANGES.rst#370-2020-10-24

Dreamsorcerer:
Appears all the tests are broken (in master), but otherwise this should be good to merge.

Also not sure why there is a CLA prompt, this is not needed for any of the other aio-libs projects that I've seen.

Dreamsorcerer:
So, a bug aiohttp 3.7 causes major issues for aiohttp-session, so this will have to wait till 3.8 now.

Dreamsorcerer:
#575

PR #456, opened 2020-05-02
Improved demo for aioredis implementation

JeremyVriens:
## What do these changes do?
These changes on the aioredis demo "reduce complexity" and use the on_startup signal to setup aiohttp_session.

### DIFF
--- a/demo/redis_storage.py
+++ b/demo/redis_storage.py
@@ -1,9 +1,8 @@
-import asyncio
 import aioredis
 import time
 
 from aiohttp import web
-from aiohttp_session import setup, get_session
+from aiohttp_session import get_session, setup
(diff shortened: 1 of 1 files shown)


Dreamsorcerer:
I'd suggest simplifying this even further to just a single cleanup context function:
```
def setup_redis_session(app):
    [start redis]
    [setup session]
    yield
    [cleanup redis]
```

It's cleaner to keep the startup/shutdown of Redis within the same function, and also safer, as it can help ensure that Redis is actually shutdown if an exception happens somewhere else during startup.

Dreamsorcerer:
I realise it's been a year, but any interest in finishing off this PR and getting it merged?

Dreamsorcerer:
Closed in #609

PR #464, opened 2020-06-01
Encode SimpleCookieStorage's cookies to be RFC-compliant

speezepearson:
## What do these changes do?

Perform %-encoding on `SimpleCookieStorage`'s JSON strings to make them obey [RFC 6265](https://tools.ietf.org/html/rfc6265.html#section-4.1.1) more strictly: the RFC says the server "should not" include certain characters in cookies (notably, spaces, commas, and quotes, which all appear plentifully in the cookie's JSON). I discovered this when I found that [HTTPie](https://httpie.org/) stored not-_quite_-right versions of the data that I was trying to store in a session -- here's a demo [before](https://pastebin.com/NHzSadAu) and [after](https://pastebin.com/mAUzj6e0) the fix.

I'm not thrilled about having hand-written the %-[en/de]coding functions. [`urllib.parse.quote`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote) would also be a reasonable approach, but AFAICT it just _happens_ to work for this purpose, rather than working _by design_: it _happens_ to %-encode `\`, even though [that's not a reserved character according to the RFC that that function is based on.](https://tools.ietf.org/html/rfc3986.html#section-2.2)

I'm also unhappy about how the Cookie header field is less readable. But that seems like a lesser evil than sending technically-non-RFC-compliant cookie values.

(I'm open to the possibility that this should be a PR against aiohttp instead -- after all, it seems like aiohttp _shouldn't let you_ store invalid cookies. But I suspect that they'll be very reluctant to make that sort of backwards-incompatible change.)

It makes the cookie less human-readable. Aside from that, there's no functional change.

## Related issue number
(none)

## Checklist

- [x] **This section of the template is out of date.** Add a new news fragment into the `CHANGES` folder


speezepearson:
Closing this: CLA-Assistant requests permissions on my account that I'm not willing to give it, and I doubt anybody values this PR enough to go through the trouble of some nonstandard licensing procedure.

PR #468, opened 2020-06-06
Custom Storage key/Redis delimiters

agentspawn:
Branch adds user-specified Redis and other storage key delimiters. People are very opinionated about the structure of their Redis keys and `_` is rarely used. Delimiter can now be specified at instantiation and defaults to `:` for Redis.

End user will not experience any difference in service.

`.feature`


asvetlov:
The PR replaces one opinionated naming convention with another opinionated non-compatible naming.
I don't see why one version is better than other but want to mention that the defaults changing breaks caching on the library upgrading.
I dont want to make them hurt, sorry.
Closing

PR #512, opened 2020-11-13
Fix test failures

Dreamsorcerer:
This fixes a couple of issues with the tests. Once aiohttp 3.8 is available, we'll see if there are any other failures that need attention.


aiohttp_session/__init__.py
```diff
@@ -154,10 +154,10 @@ async def factory(request, handler):
         except web.HTTPException as exc:
             response = exc
             raise_response = True
-        if not isinstance(response, web.StreamResponse):
+        if not isinstance(response, (web.StreamResponse, web.HTTPException)):
```
Dreamsorcerer, line 157: These will be needed for aiohttp 4, but cause no harm making the changes now.

PR #513, opened 2020-11-14
Update .travis.yml

asvetlov:
Drop Python 3.5

In general, the library should migrate to GitHub actions but small Travis config change doesn't harm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,20 +7,15 @@ services:
 - docker
 
 python:
-- 3.5
 - 3.6
+- 3.7
 - nightly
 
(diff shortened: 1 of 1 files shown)


webknjaz:
TBH after the recent FOSS cut-off, I'd rather not keep Travis CI setup any longer: https://webknjaz.me/prose/et-tu-brute-use-travis-ci-for-foss-no-more/. I'd only use it as an additional CI if a given project wants to run tests in envs unavailable elsewhere.

asvetlov:
Agree

PR #527, opened 2020-12-18
Replace Travis CI with GitHub Actions CI/CD workflows

Dreamsorcerer:
Test fixes and update CI away from Travis.

This gets the tests running successfully again.

Next step will be to update aioredis again, which will require removing the `loop` parameter from conftest.py. I'm really struggling to get this working correctly though.

Then we will need to add Python 3.8/3.9 to the tests (which are currently failing atleast partially due to aioredis). (I've opened issues to track both of these).


.travis.yml
```diff
@@ -1,46 +0,0 @@
...
-      provider: pypi
-      user: aio-libs-bot
-      password:
-        secure: >
-          hlgl7Ml+yv13wqXFnzHPQKyS0Sj45TmsvC53lRk970TJD8ehfeYSxcdt6UMmxGzGUqGGdakQKUueSFxe4+hCKo1TiabsIPsSJWFXL+RfboZbcGPUVfk0MPNnOEOLBOWnsBVzfRXfWdG6oQ+Sn+2z6WIRcOaiyIFll2ABI1tkVxciDgBRl3czx13V4MFWVEM/8P9YJA7x1srzbVISlWKf1ZOA/nk5xJMyGbsdmEV4JVwFanGCTdALHxi8+8MOwFOxecbHlCSpo2od2JeHZunysHuDt3zX84zwH80LqgUyWoTCYgb5QM213NkZHAokV5JAdPTVQJvsZMk65sELGljN36TcggpeZqDlY6pS5+j116GUHbEqLnZveFYsH2XcqhsY0jcbdzlo9KjlBSBzY0VXXsspZcUVAHIQ+o7Cv/JjfabQTYVAhS4nAZy0P0bWOWfoFZAyFlnN86zePPzx1bqCTyMwlfmeNKgCQov4+OmudgDnFwJtebk4poZHB5U4fPFslNTm7bar6H16nqQ6+/FoXDQ1kiOTSmAZDlcySnPWnVaxxRyeguVTtnPJGNWYChFd/NiGGqYeH9m0GlSDUeW+TUezOLOJgF/UxVupV+ISAcKcgbm28MlQNx8/1IVOhsmY3F199eV3briHI8E3vjO67oBBURove+97s1lK4cS56n0=
-      distributions: sdist bdist_wheel
-      on:
-        tags: true
-        all_branches: true
-        python: 3.6
```
webknjaz, line 34: @Dreamsorcerer you should also migrate publishing to PyPI. Otherwise, you'll lose the ability to make releases.
I've added a secret called `PYPI_API_TOKEN`, follow pointers at https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/ to reimplement this within GHA.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        pip install --upgrade pip
+        pip install -r requirements-dev.txt
```
webknjaz, line 22: I recommend putting separate commands into separate steps because they are timed separately and displayed better in the log this way. When something fails, it's easier to narrow down what went wrong too.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,30 @@
...
+        pip install codecov
```
webknjaz, line 23: You can use https://github.com/codecov/codecov-action and this will be unnecessary.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+on: pull_request
+
+jobs:
+  test:
+    name: Tests
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7]
```
webknjaz, line 11: I think you shouldn't lose nightlies.

Dreamsorcerer:
@webknjaz I tried to set this up the same as the other project, but seems I've broken codecov?

webknjaz:
The log suggests that the coverage file gets found and uploaded. What are you talking about exactly?

pytest.ini
```diff
@@ -1,3 +1,6 @@
 [pytest]
 filterwarnings=
     error
+    # This is internal to the docker library.
+    ignore:the imp module is deprecated:DeprecationWarning
```
webknjaz, line 5: It's best to use the fourth parameter (add yet another semicolon) to specify where exactly this exception is expected/allowed to originate from.

Makefile
```diff
@@ -11,7 +11,7 @@ vtest: flake develop
 	py.test ./tests/
 
 cov cover coverage: flake
-	py.test --cov aiohttp_session --cov-report html ./tests/
+	py.test --cov aiohttp_session --cov-report html --cov-report=xml ./tests/
```
webknjaz, line 14: It's not related to this PR but the preferred spelling of the `pytest` command is the one without the dot in the middle.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,32 @@
...
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        pip install --upgrade pip
+        pip install -r requirements-dev.txt
+    - name: Run tests
+      run: |
+        make cov
+        python setup.py check -rms
```
webknjaz, line 27: Not sure if this is still relevant. I suggest exploring (in a separate PR, of course) if all the checks embedded into this command can be replaced by `twine check`.

webknjaz (approved):

Dreamsorcerer:
> The log suggests that the coverage file gets found and uploaded. What are you talking about exactly?

I was looking at the required codecov/patch in the checks summary. Does it just need disabling?

webknjaz:
Ah, I haven't noticed that from the phone. I suppose codecov was drunk and hasn't updated the status on GH — this happens sometimes. I'll just merge this PR forcefully. If it'll keep being a problem, we'll reconsider the branch protection settings.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,32 @@
+name: Tests
+
+on: pull_request
```
webknjaz, line 3: @Dreamsorcerer I think I figured it out... So it looks like since we don't run GHA on `push`, there's nothing reporting coverage for `master` anymore which in turn may affect what coverage reports.

webknjaz: https://app.codecov.io/gh/aio-libs/aiohttp-session/ — no updated coverage after merging this PR

PR #530, opened 2020-12-20
Redis delimiter

TheDoctorAI:
This is from a [previous PR](https://github.com/aio-libs/aiohttp-session/pull/468) by someone else. I also need this feature so I merged the latest master to it for resubmit. From the original PR is commented `the defaults changing breaks caching on the library` but this was not actually true. I believe [this commit](https://github.com/aio-libs/aiohttp-session/pull/468/commits/e492b6703868b0c4ffc74efafeadbc127f5b425a) addressed that.

The reason I need this is because without it, I have to create special exceptions in a different code-base the way it is now to prevent collision. If I can set the delimiter, it saves me many hours of work.


Dreamsorcerer:
Maybe I'm just not familiar enough with Redis, but I'm struggling to see why the delimiter needs to be changed. Could you provide an example or something that shows why this change is useful?

Dreamsorcerer:
Closing as stale and a lack of clarification.

PR #555, opened 2021-03-17
Predicate method to test for active session

RichardBruskiewich:
Using `get_session()` always activates a new session, from `session = await storage.load_session(request)`. That's ok for some contexts, but not for some situations (like authentication verification contexts).

## What do these changes do?

Added a simply predicate which checks the available storage for a saved **AIOHTTP_SESSION** cookie, returning True if one Found, False otherwise.

An alternative(?) to adding this new function would be to add an optional boolean flag directly to the get_session() method, which allows creation of the new session cookie but otherwise triggers the raising of a RunTimeError if there is no session cookie is already available in the session storage. Not sure what to call such a flag - maybe `create`? -nor what its default value ought to be (perhaps **True** to maintain non-breaking backwards compatibility to library users).

Maybe both changes could be pondered although the second option seems more challenging to implement due to the embedded call to `storage.load_session(request)`.

Either way, one can test first if a session is active, then get it afterwards or otherwise make a suitable application decision to create one (or work outside such a session, e.g. for public non-authenticated access).

No, unless they use the new function. But if they have the problem with get_session() described above, then this function should be useful.

## Related issue number

n/a - this PR documents the problem

## Checklist

- [ ] I think the code is well written

### DIFF
--- a/aiohttp_session/__init__.py
+++ b/aiohttp_session/__init__.py
@@ -104,6 +104,12 @@ class Session(MutableMapping):
 SESSION_KEY = 'aiohttp_session'
 STORAGE_KEY = 'aiohttp_session_storage'
 
+async def is_active_session(request):
+    cookie = storage.load_cookie(request)
+    if cookie is not None:
+        return True
+    else:
+        return False
 
 async def get_session(request):
     session = request.get(SESSION_KEY)


aiohttp_session/__init__.py
```diff
@@ -104,6 +104,12 @@ def __delitem__(self, key):
 SESSION_KEY = 'aiohttp_session'
 STORAGE_KEY = 'aiohttp_session_storage'
 
+async def is_active_session(request):
+    cookie = storage.load_cookie(request)
+    if cookie is not None:
+        return True
+    else:
+        return False
```
webknjaz, line 112: The idiomatic way of doing this is
```suggestion
    return cookie is not None
```

aiohttp_session/__init__.py
```diff
@@ -104,6 +104,12 @@ def __delitem__(self, key):
 SESSION_KEY = 'aiohttp_session'
 STORAGE_KEY = 'aiohttp_session_storage'
 
+async def is_active_session(request):
```
webknjaz, line 107: This function needs a docstring

webknjaz:
@RichardBruskiewich this PR needs at least a docs update and a test. Please also rebase it so it'll pull in the refreshed CI setup.
Other than that, I'll leave this to @Dreamsorcerer to assess whether this function is useful and accept the PR.

aiohttp_session/__init__.py
webknjaz: Also, the name `is_session_active` would sound better.

webknjaz: Oh, and `async` is absolutely unnecessary because it's synchronous and doesn't use `await` anywhere.

webknjaz (commented):
Looking at this more, it doesn't seem to be necessary. It's not that hard to type `storage.load_cookie(request) or raise RuntimeError`.
OTOH I don't fully understand the use case — why would one use the presence of session as an auth indication? It's weird. The standard way is to just check if there's a flag in the session. It won't be there in the new sessions, only the pre-existing ones.

So I'm -1 on this change.

aiohttp_session/__init__.py
RichardBruskiewich: agreed, lol

aiohttp_session/__init__.py
RichardBruskiewich: re: asynch - absolutely right... sorry... force of habit since so many of my aiohttp related functions have been async as of necessity. Can be removed.

I'm also fine with the rename `is_session_active` although I guess I'm thinking of the conditional construct 

```
if is_active_session(request):
   ...
```

which sounds a slightly more natural English than

```
if is_session_active(request):
 ...
```

webknjaz: I think `is_active_session` sounds like the expectation is that you're checking that the argument is a session and it's weird to see a request there instead.

Dreamsorcerer: > I think `is_active_session` sounds like the expectation is that you're checking that the argument is a session and it's weird to see a request there instead.

Yep, I'd read that in English as "Is this an active session?" vs. "Is the session currently active (in this request)?".

Without context, both sound natural.

Dreamsorcerer:
I agree that it's not clear exactly what this is achieving. Some examples (with code) of where this function would be useful might help explain that.

RichardBruskiewich:
Thank you colleagues for your quick and continuing replies to the initial PR. Much more responsive than some other open source projects I know about. 

My core use case is to implement (in aiohttp) an OAuth2 driven user login/logout session management as part of a larger system. 

It is probably my relatively ignorance about `aiohttp_session `(and generally, `aiohttp`) best practices holding me back.  

I note that the `aiohttp_session` function `new_session` creates a `new`, `empty`, `changed==False` Session object in the storage context then injects it into the input `request` context.  My fresh understanding here is that the  `get_session` method is similar except that it first checks for a Session object in the request context, then, if none found, also returns a `new`, `empty`, `changed==False` Session.  

However, in both cases, nothing happens unless a user assigns the tag=value data to the Session, in which case, the middleware should automatically persist it in the storage context (generally, with a cookie).  In which case, what I'm considering an 'active' session would simply be a _non-empty_ Session object which is persisted (and tracked by the storage context using a cookie).

I'm not sure, though, that this happens during HTTP redirections, which is a challenge for the OAuth2 part of my application, which redirects in between an external 3rd party OAuth2 authority and my application.

I'll close this PR for now, but may reopen it again if there is a need to iterate further (even from a documentation point of view).

Thanks.

Dreamsorcerer:
> I'm not sure, though, that this happens during HTTP redirections, which is a challenge for the OAuth2 part of my application, which redirects in between an external 3rd party OAuth2 authority and my application.

I think you've just stumbled on to the bug that has been causing issues for everyone with aiohttp 3.7 (it's fixed in the upcoming 3.8 release). Try downgrading to 3.6 and see if that works better, or if you can't do that, you may be able to work around it temporarily by `return`ing the exception instead of `raise`ing it (though you should change back to `raise` once 3.8 is out).

webknjaz:
@RichardBruskiewich my understanding from web programming in general and frameworks in different language stacks is that you are supposed to just always use `get_session()` and set some "authenticated=True" or "user_id=666" on the session in the authentication handler and use `get_session()` + check for that flag whenever you have "if authenticated" checks. If the flag is not there, treat it as an unauthenticated session (that should work regardless of whether the session didn't exist before the current request or if it existed and the user clicked "sign off" and you set "authenticated=False").

RichardBruskiewich:
> 
> 
> > I'm not sure, though, that this happens during HTTP redirections, which is a challenge for the OAuth2 part of my application, which redirects in between an external 3rd party OAuth2 authority and my application.
> 
> I think you've just stumbled on to the bug that has been causing issues for everyone with aiohttp 3.7 (it's fixed in the upcoming 3.8 release). Try downgrading to 3.6 and see if that works better, or if you can't do that, you may be able to work around it temporarily by `return`ing the exception instead of `raise`ing it (though you should change back to `raise` once 3.8 is out).

Thanks @Dreamsorcerer for the tip. I downgraded to aiohttp 3.6 and this enabled the Session object to propagate, although I did notice that the 'identifier' field value is lost along the way.  I can, of course, record the value in another session field, and maybe this too will be fixed in aiohttp 3.8, but just mentioning.

I'll now see how far I get now.

Dreamsorcerer:
> although I did notice that the 'identifier' field value is lost along the way.

Feel free to open a bug report with some code to reproduce the issue, as I'm not really sure I understand.

RichardBruskiewich:
> 
> 
> > although I did notice that the 'identifier' field value is lost along the way.
> 
> Feel free to open a bug report with some code to reproduce the issue, as I'm not really sure I understand.

I'll recheck the exact context for this observation but it just seemed that after I set the Session identifier (alongside the Session data) in one web request, it seemed that the Session didn't retain it in the client browser's next request transaction (although the Session data made it through).

RichardBruskiewich:
> > I'm not sure, though, that this happens during HTTP redirections, which is a challenge for the OAuth2 part of my application, which redirects in between an external 3rd party OAuth2 authority and my application.
> 
> I think you've just stumbled on to the bug that has been causing issues for everyone with aiohttp 3.7 (it's fixed in the upcoming 3.8 release). Try downgrading to 3.6 and see if that works better, or if you can't do that, you may be able to work around it temporarily by `return`ing the exception instead of `raise`ing it (though you should change back to `raise` once 3.8 is out).

Good afternoon @Dreamsorcerer , do you have any idea when the "_upcoming AIOHTTP 3.8 release_" will actually be released?  How big is the patch to fix the issue you acknowledged? Could it be issued as a bug fix to 3.7.3 => 3.7.4?

RichardBruskiewich:
> > I'm not sure, though, that this happens during HTTP redirections, which is a challenge for the OAuth2 part of my application, which redirects in between an external 3rd party OAuth2 authority and my application.
> 
> I think you've just stumbled on to the bug that has been causing issues for everyone with aiohttp 3.7 (it's fixed in the upcoming 3.8 release). Try downgrading to 3.6 and see if that works better, or if you can't do that, you may be able to work around it temporarily by `return`ing the exception instead of `raise`ing it (though you should change back to `raise` once 3.8 is out).

@asvetlov, do you have any idea when aiohttp 3.8 or better, will be released?

PR #569, opened 2021-04-04
Expand typing and general tidyup

Dreamsorcerer:


Dreamsorcerer:
@webknjaz The codecov/patch check gives a 403 when I click for more details. Can we fix that, or disable the requirement?

Dreamsorcerer:
@webknjaz Can we remove the requirement on codecov/patch? I think we should be able to make judgement calls on this.

webknjaz:
@Dreamsorcerer maybe you could just lower the threshold yourself? https://docs.codecov.io/docs/codecov-yaml

Dreamsorcerer:
Sure, I'll give that a go.

PR #575, opened 2021-04-25
Add samesite support.

### LINKED ISSUE #536: aiohttp 3.7 samesite flag voor cookies cant be set
since aiohttp 3.7, there is now an option to use samesite flag (which defaults to None) for cookies.
However aiohttp-session has no way to pass this

Dreamsorcerer:
Pending aiohttp 3.8 release.

Fixes #536

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ with open(os.path.join(os.path.abspath(os.path.dirname(
 def read(f):
     return open(os.path.join(os.path.dirname(__file__), f)).read().strip()
 
-install_requires = ['aiohttp>=3.0.1', 'typing_extensions>=3.7.4; python_version<"3.8"']
+install_requires = ['aiohttp>=3.8', 'typing_extensions>=3.7.4; python_version<"3.8"']
 extras_require = {
     'aioredis': ['aioredis>=1.0.0'],
     'aiomcache': ['aiomcache>=0.5.2'],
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -14,7 +14,7 @@
 def read(f):
     return open(os.path.join(os.path.dirname(__file__), f)).read().strip()
 
-install_requires = ['aiohttp>=3.0.1', 'typing_extensions>=3.7.4; python_version<"3.8"']
+install_requires = ['aiohttp>=3.8', 'typing_extensions>=3.7.4; python_version<"3.8"']
```
kellerza, line 17: Do we really need 3.8, or can we get away with [3.7.0](https://github.com/aio-libs/aiohttp/releases/tag/v3.7.0)?

```suggestion
install_requires = ['aiohttp>=3.7.0', 'typing_extensions>=3.7.4; python_version<"3.8"']
```

Dreamsorcerer: If we could use 3.7, I wouldn't have specified 3.8. :P
3.7 has a regression that breaks core functionality in aiohttp-session, and thus also fails a bunch of tests (look around for the dozen or more mentions in various bug reports about the problem).

Dreamsorcerer:
I need to resolve mypy for aiohttp 3.8 in a separate PR. Will do later in the week.

Dreamsorcerer:
Nevermind, just needed to merge master.

kellerza:
Seems like this just missed the first 2.10 beta - https://github.com/aio-libs/aiohttp-session/blob/v2.10.0a0/aiohttp_session/__init__.py#L214

PR #601, opened 2021-07-13
New EncryptedRedisStorage.

RomaLash:
## New EncryptedRedisStorage.

Added possibilities to delete key from redis and encrypt cookie data.

## Checklist

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -88,6 +88,13 @@ Available session storages are:
 
       $ pip install aiohttp_session[aioredis]
 
+* ``aiohttp_session.redis_storage.EncryptedRedisStorage(redis_pool, secret_key)`` -- stores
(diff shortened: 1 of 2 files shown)


Dreamsorcerer:
I don't understand what benefit the encryption is meant to add?
The cookie in the redis storage is just a key, it doesn't contain any data...

I'm also not sure what adding `delete_session()` achieves either. This does not seem to appear anywhere in the aiohttp-session API, so it would not get used. If we wanted to add an API to delete a session, it would need to be included as part of the public API (and should be done in a separate PR).

Dreamsorcerer:
Sorry it's taken over a year to look at, but currently this doesn't seem to add anything, so I'm closing this. Feel free to continue the discussion though, and we can reopen if it's still relevant.

PR #611, opened 2021-08-07
Upgrade to aioredis 2

### LINKED ISSUE #528: Update aioredis
Tests are currently failing on latest aioredis. This will require updating conftest.py to remove the `loop` parameter.

Dreamsorcerer:
Fixes #528.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -104,7 +104,6 @@ def redis_server(  # type: ignore[misc]  # No docker types.
                 aioredis.from_url("redis://{}:{}".format(host, port))  # type: ignore[no-untyped-call]  # noqa: B950
             )
             loop.run_until_complete(conn.execute('SET', 'foo', 'bar'))
-            await conn.execute('SET', 'foo', 'bar')
             break
         except ConnectionRefusedError:
             time.sleep(delay)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -100,8 +100,7 @@ def redis_server(  # type: ignore[misc]  # No docker types.
     delay = 0.1
     for _i in range(20):
         try:
-            conn = loop.run_until_complete(_create_redis(host, port))
-            loop.run_until_complete(conn.execute('SET', 'foo', 'bar'))
+            loop.run_until_complete(_populate_redis(host, port))
             break
         except ConnectionRefusedError:
             time.sleep(delay)
@@ -201,5 +200,6 @@ def memcached(  # type: ignore[misc]
     conn.close()
 
 
-async def _create_redis(host: str, port: int) -> aioredis.Redis:
-    return await aioredis.from_url("redis://{}:{}".format(host, port))  # type: ignore[no-untyped-call]  # noqa: B950
+async def _populate_redis(host: str, port: int) -> None:
+    redis = await aioredis.from_url("redis://{}:{}".format(host, port))  # type: ignore[no-untyped-call,no-any-return]  # noqa: B950
+    await redis.set("foo", "bar")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -40,7 +40,7 @@ def unused_port() -> int:
 @pytest.fixture(scope="session")
 def loop() -> Iterator[asyncio.AbstractEventLoop]:
     loop = asyncio.new_event_loop()
-    asyncio.set_event_loop(loop)
+    asyncio.set_event_loop(None)
 
     yield loop
 
@@ -100,9 +100,7 @@ def redis_server(  # type: ignore[misc]  # No docker types.
     delay = 0.1
     for _i in range(20):
         try:
-            conn = loop.run_until_complete(
-                aioredis.from_url("redis://{}:{}".format(host, port))  # type: ignore[no-untyped-call]  # noqa: B950
-            )
+            conn = loop.run_until_complete(create_redis())
             loop.run_until_complete(conn.execute('SET', 'foo', 'bar'))
             break
         except ConnectionRefusedError:
@@ -201,3 +199,7 @@ def memcached(  # type: ignore[misc]
     conn = aiomcache.Client(**memcached_params)
     yield conn
     conn.close()
+
+
+async def _create_redis() -> aioredis.Redis:
+    return await aioredis.from_url("redis://{}:{}".format(host, port))  # type: ignore[no-untyped-call]  # noqa: B950
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -196,6 +196,7 @@ def memcached_params(memcached_server: _ContainerInfo) -> _MemcachedParams:  # t
 
 @pytest.fixture
 def memcached(  # type: ignore[misc]
+    loop: asyncio.AbstractEventLoop,
     memcached_params: _MemcachedParams
 ) -> Iterator[aiomcache.Client]:
     conn = aiomcache.Client(loop=loop, **memcached_params)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -100,7 +100,10 @@ def redis_server(  # type: ignore[misc]  # No docker types.
     delay = 0.1
     for _i in range(20):
         try:
-            loop.run_until_complete(_populate_redis(host, port))
+            conn = loop.run_until_complete(
+                aioredis.from_url("redis://{}:{}".format(host, port))  # type: ignore[no-untyped-call]  # noqa: B950
+            )
+            loop.run_until_complete(conn.set("foo", "bar"))
             break
         except ConnectionRefusedError:
             time.sleep(delay)
@@ -198,8 +201,3 @@ def memcached(  # type: ignore[misc]
     conn = aiomcache.Client(**memcached_params)
     yield conn
     conn.close()
-
-
-async def _populate_redis(host: str, port: int) -> None:
-    redis = await aioredis.from_url("redis://{}:{}".format(host, port))  # type: ignore[no-untyped-call,no-any-return]  # noqa: B950
-    await redis.set("foo", "bar")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -129,6 +129,7 @@ def redis(
     async def start(pool: aioredis.ConnectionPool) -> aioredis.Redis:
         return aioredis.Redis(connection_pool=pool)
 
+    asyncio.set_event_loop(loop)
     pool = aioredis.ConnectionPool.from_url(redis_url)
     redis = loop.run_until_complete(start(pool))
     yield redis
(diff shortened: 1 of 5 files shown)


Dreamsorcerer:
@webknjaz Do you have any idea what the error in CI means? I tried removing the loop() fixture in conftest.py in the hope of fixing other issues, but now it's complaining about a loop() fixture in aiohttp's plugin...

Specific changes that introduced this error: https://github.com/aio-libs/aiohttp-session/pull/611/files/9bf3ff0e3e8f2800c5dc0d39525f106009ad1ca5..5f0ffbd7932e2ee256e0fb0ae4cdcad50d73cddc

tests/conftest.py
```diff
@@ -69,11 +48,10 @@ def docker() -> DockerClient:  # type: ignore[misc]  # No docker types.
 
 
 @pytest.fixture(scope='session')
-def redis_server(  # type: ignore[misc]  # No docker types.
+async def redis_server(  # type: ignore[misc]  # No docker types.
```
webknjaz, line 51: This change is likely the reason of the problem — pytest fixtures are synchronous AFAIK.

Dreamsorcerer: Hmm, this suggested that fixtures could be async:
https://stackoverflow.com/questions/49936724/async-fixtures-with-pytest#49940520

webknjaz: I think it may depend on what pytest plugins get used. Anyway, this should give you an idea of where to start bisecting the issue.

webknjaz: Fixtures that pytest-aiohttp provides usually rely on having a disposable event loop where those tests will run. If a fixture is a coroutine on the outer pytest level, I suspect it may run in a different event loop and I'm not ready to guess what the consequences of that would be.

webknjaz: FWIW try to make this work w/o this part of the refactoring and see if that helps.

tests/conftest.py
```diff
@@ -37,21 +35,6 @@ def unused_port() -> int:
     return port
 
 
-@pytest.fixture(scope='session')
-def loop() -> Iterator[asyncio.AbstractEventLoop]:
-    loop = asyncio.new_event_loop()
```
webknjaz, line 46: FYI removing the per-test loop may mean that tests mutating it may influence each other depending on a shared state.

tests/conftest.py
Dreamsorcerer: It's not refactoring, it's just trying to get the thing working. :P
I'll revert these changes and take another look at the previous errors.

Dreamsorcerer: Yep, so now back to `There is no current event loop in thread 'MainThread'.`
As the `loop` parameter has been removed from just about everything now, I don't know how to make this work.
I'm not sure why `loop.run_until_complete()` isn't able to set itself as the current event loop for the length of the run.

Dreamsorcerer: I tried changing `asyncio.set_event_loop(loop)` instead of `None`. But, that hasn't made a difference.

webknjaz: I think that maybe when there's `loop.run_until_complete(do_smth())`, that `do_smth()` triggers the problem. Maybe try wrapping it into another coroutine and see what happens?
As in
```python
async def coro():
    return await do_smth()

loop.run_until_complete(coro())
```

webknjaz: Try running just one test with `--pdb` maybe? Or inject a `breakpoint()` right before?

Dreamsorcerer: Didn't help. I don't have the docker stuff setup to run the full tests locally, but I commented out the docker parts and have just found that it successfully runs past the point of failure in CI. So, I can't even reproduce the exception locally...

Dreamsorcerer: I'm feeling like this is something @asvetlov would be able to figure out?

panagiks: My 2 cents: The loop fixture sets the event loop to `None` in its cleanup [here](https://github.com/aio-libs/aiohttp-session/blob/1a589e37371c35b52d637035e54827db6a175c55/tests/conftest.py#L52) and setup [here](https://github.com/aio-libs/aiohttp-session/blob/1a589e37371c35b52d637035e54827db6a175c55/tests/conftest.py#L43). Reading the error, `aioredis` tries to acquire an `asyncio.Lock` which in turn internally runs `events.get_event_loop()` which since the `loop` fixture set the running loop to `None` fails.

Suggestion:
try setting the global loop to the fixture loop inside the redis_server fixture (i.e. add [here](https://github.com/aio-libs/aiohttp-session/blob/1a589e37371c35b52d637035e54827db6a175c55/tests/conftest.py#L74) `asyncio.set_event_loop(loop)`). This is bad practice IMO (so it wouldn't be a permanent solution) but if this error is then fixed it will tell us we are at least in a good path.

Sorry for not testing it myself ATM but I only have python 3.8 on my work machine (and the tests already fail on 3.8). I'll try to offer a bit more hands-on help during the weekend (but tentative if I'll have time).

panagiks: Also, [this](https://aioredis.readthedocs.io/en/latest/migration/#connecting-to-redis) suggests that `aioredis.from_url` is not a coro.

panagiks: Minimum reproducible example:

```py
>>> import asyncio
>>> import aioredis
>>> r = aioredis.from_url("redis://{}:{}".format("foo", 123))
>>> asyncio.set_event_loop(None)
>>> r = aioredis.from_url("redis://{}:{}".format("foo", 123))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/panagiks/.venvs/aioredis2/lib/python3.8/site-packages/aioredis/utils.py", line 25, in from_url
    return Redis.from_url(url, **kwargs)
  File "/home/panagiks/.venvs/aioredis2/lib/python3.8/site-packages/aioredis/client.py", line 814, in from_url
    connection_pool = ConnectionPool.from_url(url, **kwargs)
  File "/home/panagiks/.venvs/aioredis2/lib/python3.8/site-packages/aioredis/connection.py", line 1245, in from_url
    return cls(**kwargs)
  File "/home/panagiks/.venvs/aioredis2/lib/python3.8/site-packages/aioredis/connection.py", line 1270, in __init__
    self._lock = asyncio.Lock()
  File "/usr/lib/python3.8/asyncio/locks.py", line 164, in __init__
    self._loop = events.get_event_loop()
  File "/usr/lib/python3.8/asyncio/events.py", line 639, in get_event_loop
    raise RuntimeError('There is no current event loop in thread %r.'
RuntimeError: There is no current event loop in thread 'MainThread'.
```

panagiks: As for why the loop in not set inside `loop.run_until_complete` the answer is because it is never executed. Since `aioredis.from_url` is a synchronous function, it is executed before `loop.run_until_complete` and its result would have been passed to it as an argument (like running `loop.run_until_complete(str(42))` the `str(42)` would be executed first and then `loop.run_until_complete("42")` would run). Sorry for the multi-message spam :( (new ideas coming in just as I hit 'Comment') That's my last message for now :D

Dreamsorcerer: Thanks for looking at this.

> try setting the global loop to the fixture loop inside the redis_server fixture
I already tried setting the loop in the loop fixture in a previous commit. I've tried setting it in the redis_server as you suggested, and I think it's working! We've got some new errors now, so I'll try and figure those out now.

> Sorry for not testing it myself ATM but I only have python 3.8 on my work machine (and the tests already fail on 3.8).
Yep, this is first step towards getting 3.8+ running.

> Also, this suggests that aioredis.from_url is not a coro.
Yes, seems a little weird. It returns the client, then an `await`, `async with` or simply executing a command all do the same thing and initialise the client.

Dreamsorcerer: OK, looks like it's working now. If you've got any ideas on a better way to do it, that would be great. It looks like we have the same issue with aiomcache when we remove the loop parameter from that.

We are also getting connection errors on about half the CI runs, requiring them to be rerun repeatedly. Any ideas to fix that would be great as well:
`aioredis.exceptions.ConnectionError: Error 111 connecting to 172.17.0.3:6379. 111.`

Dreamsorcerer:
@webknjaz Can we make codecov/patch optional? This is the 2nd time it's blocked me. It's complaining about 3 lines of untested code which have only had comments/annotations changed.

tests/conftest.py
webknjaz: Well, maybe implement an exponential backoff? Or a fixture that would wait for the service to be actually up and running before proceeding with tests?

webknjaz:
> @webknjaz Can we make codecov/patch optional?

How about patching `codecov.yml` to be less sensitive to small changes?

Dreamsorcerer:
> > @webknjaz Can we make codecov/patch optional?
> 
> How about patching `codecov.yml` to be less sensitive to small changes?

I don't see anything in the documentation that might do this. It's also not really the size, but the fact that it doesn't know the difference between functional changes and style/comment changes. It just seems rather awkward trying to hack the config to convince codecov to pass. Code coverage is useful as a guiding tool, I don't see it being useful as a hard rule that can't be bypassed when we know better (and we don't require it on the aiohttp repo...).

Dreamsorcerer:
On aiohttp you enabled an option to skip the chronographer with a label. If that's easy to setup for codecov as well, I'd be happy with that.

webknjaz:
> On aiohttp you enabled an option to skip the chronographer with a label. If that's easy to setup for codecov as well, I'd be happy with that.

Chronographer is an app that I wrote so I implemented this in its code. We don't have such a level of control over Codecov. But I'm pretty sure it's possible to make it less sensitive.

webknjaz:
@Dreamsorcerer you probably need to adjust this https://docs.codecov.com/docs/commit-status#branches

Dreamsorcerer:
> @Dreamsorcerer you probably need to adjust this https://docs.codecov.com/docs/commit-status#branches

The branches option? Wouldn't that just result in it not reporting anything for some branches? That would make it impossible to merge as it is a required check.

There are other options there like `informational` which would just make the check pass regardless, but that's fundamentally the same as removing the required status from Github, except it would make it harder for maintainers to notice when it should be failing.

webknjaz:
Ah, sorry, I somehow copied the wrong link. Take a look at the patch status options: https://docs.codecov.com/docs/commit-status#patch-status.

Dreamsorcerer:
> Ah, sorry, I somehow copied the wrong link. Take a look at the patch status options: https://docs.codecov.com/docs/commit-status#patch-status.

I still don't see anything useful there...
I also think (unless they learn to filter out cosmetic changes) that this should have failed, and therefore alerted the maintainer. The maintainer can then make an informed decision about whether there needs to be a new test or not.

webknjaz:
I don't think so. If a patch adds uncovered lines, it should catch this. I think that setting a threshold of 75% would make it possible.

webknjaz:
FWIW I dropped the patch requirement for now.

tests/conftest.py
Dreamsorcerer: I'll follow up these changes in another PR. @panagiks Let me know if you come up with any improvements.

PR #613, opened 2021-08-19
Use C locale in expires param

exstrim401:
## What do these changes do?

Fix incorrect expires when using languages that differ from english.
Safari was deleting cookies after restart if there is incorrect expires

## Checklist

- [X] I think the code is well written


Dreamsorcerer:
It's going to atleast need a test to reproduce the original issue.

It also looks like the original formatting is a little off: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date
Appears there should be spaces instead of `-`. Plus `%T` doesn't appear to be standardised, so it'd probably be safer to do something more like the Python example: `"%a, %d %b %Y %H:%M:%S GMT"`

aiohttp_session/__init__.py
```diff
@@ -17,6 +19,15 @@
     from typing_extensions import TypedDict
 
 
+@contextmanager
+def setlocale(name):
+    saved = locale.setlocale(locale.LC_ALL)
```
Dreamsorcerer, line 24: Would it also work correctly if we only modified `locale.LC_TIME`?

PR #657, opened 2021-12-31
Use release environment for deploy

Dreamsorcerer:
@webknjaz Do you want to setup the release env? I've tested this on aiohttp-devtools and it works.

Although, worth noting that I ran the deploy in the wrong environment first time and the CI says the deploy was a success despite it not uploading to pyPI.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,6 +36,7 @@ jobs:
 
   deploy:
     name: Deploy
+    environment: release
     if: github.event_name == 'push' && contains(github.ref, 'refs/tags/')
     needs: [test]
     runs-on: ubuntu-latest


webknjaz:
> @webknjaz Do you want to setup the release env? I've tested this on aiohttp-devtools and it works.

Configured.

> Although, worth noting that I ran the deploy in the wrong environment first time and the CI says the deploy was a success despite it not uploading to pyPI.

Not sure what you mean by that.

Dreamsorcerer:
> > Although, worth noting that I ran the deploy in the wrong environment first time and the CI says the deploy was a success despite it not uploading to pyPI.
> 
> Not sure what you mean by that.

It appears that if the token is not set, then the GHA is reported as a success even though no release was pushed to PyPI.

webknjaz:
> It appears that if the token is not set, then the GHA is reported as a success even though no release was pushed to PyPI.

I don't think that's the case but maybe this is happening when there are no dists? Do you have a link to the workflow run log for me to check?

Dreamsorcerer:
> Do you have a link to the workflow run log for me to check?

This one appears successful, but I forgot the environment, so it actually never uploaded. You'll see the next commit after this one, I add in the environment parameter.
https://github.com/aio-libs/aiohttp-devtools/runs/4674771764?check_suite_focus=true

webknjaz:
Oh, that one uses Andrew's action, not mine. And it just skips the upload step if there's no token 🤷‍♂️. I was talking about how the PyPA action works.
One thing to note is that I also expected the deploy step to depend on the check.

PR #663, opened 2022-01-20
Drop Python 3.6

Dreamsorcerer:


Dreamsorcerer:
@asvetlov Can you update the required status? Have added the check thing like other repos.

PR #701, opened 2022-05-15
Replace aioredis with redis 4.3 or higher

achimnol:
## What do these changes do?

It replaces aioredis v2 dependency with redis v4.3 or higher,
in favor of official aioredis inclusion in the redis-py project since v4.2.

Redis storage users should update their aioredis v2 dependency to redis v4.3.

## Related issue number

This is a standalone PR.

## Checklist

- [x] Add a new news fragment into the `CHANGES` folder


achimnol:
Reported spurious type check failures as python/typeshed#7837.

aiohttp_session/redis_storage.py
```diff
@@ -1,14 +1,14 @@
...
-from distutils.version import StrictVersion
 from typing import Any, Callable, Optional
 
 from aiohttp import web
 
 from . import AbstractStorage, Session
 
 try:
-    import aioredis
+    import redis
```
Dreamsorcerer, line 10: I'd consider just doing `from redis import __version__ as redis_version`. Which will also avoid the lgtm warning.

aiohttp_session/redis_storage.py
```diff
@@ -43,13 +43,14 @@ def __init__(
             decoder=decoder,
         )
         if aioredis is None:
-            raise RuntimeError("Please install aioredis")
+            raise RuntimeError("Please install redis")
         # May have installed aioredis separately (without aiohttp-session[aioredis]).
-        if StrictVersion(aioredis.__version__).version < (2, 0):
-            raise RuntimeError("aioredis<2.0 is not supported")
+        lib_version = tuple(map(int, redis.__version__.split('.')[:2]))  # type: ignore
+        if lib_version < (4, 3):
```
Dreamsorcerer, line 49: Appears to already be a `redis.VERSION`, so there's no need to duplicate the tuple code here.

Dreamsorcerer: Also, adding to stubs: https://github.com/python/typeshed/pull/7838

aiohttp_session/redis_storage.py
```diff
@@ -43,13 +43,14 @@ def __init__(
             decoder=decoder,
         )
         if aioredis is None:
-            raise RuntimeError("Please install aioredis")
+            raise RuntimeError("Please install redis")
         # May have installed aioredis separately (without aiohttp-session[aioredis]).
-        if StrictVersion(aioredis.__version__).version < (2, 0):
-            raise RuntimeError("aioredis<2.0 is not supported")
+        lib_version = tuple(map(int, VERSION[:2]))
```
Dreamsorcerer, line 48: This is either already an int, or this code will error. So, if we're confident they are ints, we should just do a `cast()`.

Dreamsorcerer:
We'll need to fix up the remaining type errors, which might require some more changes to typeshed. I'll take a look at it eventually, if you don't get to it first.

achimnol:
@Dreamsorcerer Could you review this again?

amenezes:
@achimnol and @Dreamsorcerer,

Is there anything that I can make to help conclude this PR?

Dreamsorcerer:
I'll get back to it in the coming weeks, but the tests are failing in current PRs, so we'll need to update and fix those problems as the initial priority (Think it was caused by missing the required check for dependabot, so it managed to merge some broken PRs).

aiohttp_session/redis_storage.py
```diff
@@ -43,13 +43,14 @@ def __init__(
             decoder=decoder,
         )
         if aioredis is None:
-            raise RuntimeError("Please install aioredis")
+            raise RuntimeError("Please install redis")
         # May have installed aioredis separately (without aiohttp-session[aioredis]).
-        if StrictVersion(aioredis.__version__).version < (2, 0):
-            raise RuntimeError("aioredis<2.0 is not supported")
+        lib_version = VERSION[:2]
+        if lib_version < (4, 3):
```
Dreamsorcerer, line 49:
```suggestion
        if VERSION < (4, 3):
```

tests/test_redis_storage.py
```diff
@@ -40,23 +42,24 @@ async def make_cookie(
...
     cookies = client.session.cookie_jar.filter_cookies(client.make_url("/"))
     key = cookies["AIOHTTP_SESSION"]
-    encoded = await redis.get("AIOHTTP_SESSION_" + key.value)
-    s = encoded.decode("utf-8")
-    value = json.loads(s)
+    value_bytes = await redis.get("AIOHTTP_SESSION_" + key.value)
+    if value_bytes is None:
+        return None
+    value = json.loads(value_bytes)
     return value
```
Dreamsorcerer, line 58:
```suggestion
    return None if value_bytes is None else json.loads(value_bytes)
```

Dreamsorcerer:
OK, someone has sorted out the tests already, had a moment to go through this PR again. Looking good, just a couple of tweaks I'd suggest.

PR #710, opened 2022-06-20
Update __init__.py

duongtuan303030:
fix for some one like to add middlewares task like [#6789](https://github.com/aio-libs/aiohttp/issues/6789)

## What do these changes do?

 [#6789](https://github.com/aio-libs/aiohttp/issues/6789)

 [#6789](https://github.com/aio-libs/aiohttp/issues/6789)

## Checklist

- [ ] I think the code is well written


Dreamsorcerer:
Afraid this won't be accepted, as it's pretty standard to expect the `setup()` call to append and will mess up other user's who are already depending on the order.

Maybe we should deprecate/discourage the use of `Application(middlewares=...)`, so users must consider the ordering more directly...

PR #943, opened 2023-11-12
Update examples for aiohttp 3.9

JCHacking:
Edited the login_required example, the rest does not seem to need any changes


requirements.txt
```diff
@@ -1,5 +1,5 @@
 -e .
-aiohttp==3.8.5
+aiohttp==3.9.0b1
```
Dreamsorcerer, line 2:
```suggestion
aiohttp==3.9.0
```

PR #1037, opened 2024-11-25
Fix CI: update isort and flake8 for pre-commit and format files

paul-nameless:
## What do these changes do?

* Updates isort and flake8 to fix `pre-commit run`
* Fix flake8 error: long lines, unused imports, etc.
* Fix formatting of some files

Should not be any changes for the user

## Related issue number

-

## Checklist

- [X] I think the code is well written
- [X] Unit tests for the changes exist
- [X] Documentation reflects the changes

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -16,6 +16,7 @@ per-file-ignores =
     examples/*:I900
     # I900: flake8-requirements doesn't see aioredis in the requirements file cause it is extra dependency
     aiohttp_session/redis_storage.py:I900
+    build/*:I900
 
 # flake8-import-order
 import-order-style = pycharm
--- a/.flake8
+++ b/.flake8
@@ -14,6 +14,8 @@ per-file-ignores =
     tests/*:S101
     # I900: Requirements for examples shouldn't be included
     examples/*:I900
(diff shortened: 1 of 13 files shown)


tests/test_nacl_storage.py
```diff
@@ -49,7 +44,7 @@ def create_app(
 
 
 def decrypt(secretbox: nacl.secret.SecretBox, cookie_value: str) -> Any:
-    assert type(cookie_value) == str
+    assert isinstance(cookie_value, str)
```
bdraco, line 47: This is a functional change as it will now allow subclassed `str`

bdraco:
Thanks!  This looks good once the two small nits above are addressed.

bdraco (approved):

.flake8
```diff
@@ -14,6 +14,9 @@ per-file-ignores =
     tests/*:S101
     # I900: Requirements for examples shouldn't be included
     examples/*:I900
+    # I900: flake8-requirements doesn't see aioredis in the requirements file cause it is extra dependency
+    aiohttp_session/redis_storage.py:I900
```
Dreamsorcerer, line 18: I suspect it should be added to the requirements file for our testing. Though aioredis is archived, so we should also migrate to redis/valkey now.

PR #1047, opened 2024-12-18
Port to Sphinx 8.0

cjwatson:
## What do these changes do?

The old `intersphinx_mapping` format has been removed; it must now map identifiers to (target, inventory) tuples.

No.

## Checklist

- [N/A] Unit tests for the changes exist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -341,10 +341,10 @@ texinfo_documents = [
 
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {
-    "https://docs.python.org/3": None,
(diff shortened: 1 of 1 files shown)


asvetlov:
Should we back port it to 3.12?

cjwatson:
I don't think I really have an opinion - I've just been going round patching a big pile of Debian packages that are affected by this and sending things upstream where appropriate.  If you're actively supporting your 3.12 branch then it would probably make sense so that people can keep on using it with current Sphinx, and it should be pretty harmless.

Dreamsorcerer:
> Should we back port it to 3.12?

I think you're looking at the wrong repo..

